PR #3, opened 2020-07-31
Ingest service usage and new Jest test matcher

austinkelleher:


test/jest.ts
```diff
@@ -0,0 +1,175 @@
...
+}
+
+export interface GraphObjectSchema extends dataModel.IntegrationEntitySchema {
+  $schema?: string;
+  $id?: string;
+  description?: string;
+  additionalProperties?: boolean;
+}
+
+export function toMatchGraphObjectSchema<T extends Entity>(
```
austinkelleher, line 111: New custom Jest test matcher. My plan is to move this out into the SDK testing package. It reads the input schema and walks up the tree of `$ref` to build out a custom schema from the data-model.

```ts
expect(context.jobState.collectedEntities).toMatchGraphObjectSchema(...)
```

Sample usage here: https://github.com/JupiterOne/graph-google-cloud/pull/3/files#diff-1cd8a00fc9a781f92bccf7662cd6afa3R40

ndowmon: This looks awesome! Thanks for making the matcher and all the schema generation code. 

I think there are definitely arguments for / against the following, but wondering what you think of redefining the interface to:

```
function  toMatchGraphObjectSchema<T extends Entity>(
  received: T | T[],
  class: string | string[],
  expectedMatchSchema?: GraphObjectSchema, // maybe name this "overridenSchema"?
) { ... }
```

This may simplify the usage for developers for whom they often just want to validate the schema against the class(es) that they are explicitly setting. At first glance I wonder if that basic case is going to be the most common. If so, it'd simplify tests from 

```
expect(entity).toMatchGraphObjectSchema({ properties: { _class: { const: { ['Service'] } } } });  
```

to

```
expect(entity).toMatchGraphObjectSchema('Service`);
```

If most devs will be overriding, then forcing them to write the schema overrides is probably better since they'll need to be able to specify this anyway. Otherwise, a simplified interface might be best.

aiwilliams: Thanks for advancing the approach to testing!

As I understand this, it allows us to validate against the class "hierarchy" of `Service`, but also to expand the assertion with integration-specific converter verification? That is, folks could use @ndowmon 's suggested simplified interface when there are no more specific things to check? Or, is there always a need to specify the schema to ensure the `_type` value is an exact match (for example) while NOT specifying the values for other properties, only ensuring they are present and of a specific type (necessary for properties that are not already on the `Service` schema from the data model)?

austinkelleher: My vision for this would be to validate the "extended schemas" that users are creating for each new entity that they are trying to ingest. I don't think a common use case in an integration would be to validate against only the "hierarchy." If that were the case, they could just not specify any additional properties. Only specify the classes that they want to validate against.

I also considered the interface that @ndowmon suggested where the class(es) are explicitly passed to `toMatchGraphObjectSchema`. I agree that having to specify `const` on every schema declaration may be tedious and unnecessary.

austinkelleher:
> This is looking great! It looks like we'll ingest all enabled services. Does the API answer disabled services as well?

@aiwilliams Yes, I wasn't sure if we cared to ingest every API service, but perhaps we do. I already have code that supports the ability to fetch all services. It would be a simple function name change. I'm leaning towards ingesting all of the API services and not just the enabled ones...

aiwilliams:
> I wasn't sure if we cared to ingest every API service, but perhaps we do.

It would depend on the value of `find google_cloud_api_service with enabled=false`. Seems someone could live without that, and instead alert when `find google_cloud_api_service with enable=true, name="service.i.dont.want.enabled"`. That is, we don't need disabled services, because we can know they're disabled when they're not in the graph as enabled. However, folks might be confused when services aren't represented. So, I lean toward thinking we should include them, and then there would be no guessing about what is going on.

PR #6, opened 2020-08-05
Ingest IAM roles

austinkelleher:


src/getStepStartStates.ts
```diff
@@ -10,13 +11,18 @@ import { STEP_CLOUD_FUNCTIONS } from './steps/functions';
...
 
 async function getEnabledServiceNames(
   config: IntegrationConfig,
 ): Promise<string[]> {
   const client = new ServiceUsageClient({ config });
   const enabledServices = await client.collectEnabledServices();
-  return enabledServices.map((v) => v.name as string);
+  return enabledServices.map((v) => {
+    // Each value looks like this: `projects/PROJ_ID_NUM/services/appengine.googleapis.com`
+    const serviceParts = (v.name as string).split('/');
```
aiwilliams, line 23: What do you think about using a regexp to extract the part you want, and when it doesn't match, complain?

src/steps/iam/index.test.ts
```diff
@@ -0,0 +1,62 @@
...
+        properties: {
+          _type: { const: 'google_cloud_iam_role' },
+          _rawData: {
+            type: 'array',
+            items: { type: 'object' },
+          },
+          description: { type: 'string' },
+          stage: { type: 'string' },
+          custom: { const: true },
+          deleted: { type: 'boolean' },
```
aiwilliams, line 52: What does this represent? I didn't notice it in the raw data.

austinkelleher: Additional property on the role. We didn't have any roles that are in the deleted state though.

PR #10, opened 2020-08-11
Fixes #9 - Build relationship between IAM roles and users

### LINKED ISSUE #9: Build relationship between IAM roles and users
`google_cloud_iam_user` `IMPLEMENTS` `google_cloud_iam_role`
`google_cloud_service_account` `IMPLEMENTS` `google_cloud_iam_role`

austinkelleher:
#9


src/steps/resource-manager/index.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    expect(iamServiceAccountEntities.length).toBeGreaterThanOrEqual(1);
+    expect(iamServiceAccountKeyEntities.length).toBeGreaterThanOrEqual(1);
+    expect(
+      iamServiceAccountImplementsRoleRelationships.length,
+    ).toBeGreaterThanOrEqual(1);
+    expect(iamServiceAccountHasKeyRelationships.length).toBeGreaterThanOrEqual(
+      1,
+    );
+
+    expect(userEntities).toMatchGraphObjectSchema({
```
aiwilliams, line 76: I wonder if this function should fail when the collection has nothing in it, so that you don't have to make all the assertions about quantity greater or equal to 1. 🤔

src/steps/iam/__snapshots__/index.test.ts.snap
```diff
@@ -374,124 +435,166 @@ Object {
       "type": "USER_MANAGED",
       "webLink": "https://console.cloud.google.com/iam-admin/serviceaccounts/details/116674556692942636222?orgonly=true&project=j1-gc-integration-dev&supportedpurview=organizationId",
     },
+    Object {
+      "_class": Array [
+        "User",
+      ],
+      "_key": "project:j1-gc-integration-dev-user:545240943112-compute@developer.gserviceaccount.com",
```
aiwilliams, line 442: Another case where I wonder if the `project:j1-gc-integration-dev-user:` prefix is necessary. Is the email not unique across all projects?

src/steps/iam/converters.ts
```diff
@@ -62,6 +78,43 @@ function getServiceAccountWebLink({
   return `https://console.cloud.google.com/iam-admin/serviceaccounts/details/${serviceAccountId}?orgonly=true&project=${projectId}&supportedpurview=organizationId`;
 }
 
+export function toIamUserEntityKey(params: {
+  projectId: string;
+  member: string;
```
aiwilliams, line 83: How do you feel about `memberIdentifier` here?

src/steps/iam/converters.ts
```diff
@@ -13,14 +13,28 @@ import {
...
 import { generateEntityKey } from '../../utils/generateKeys';
+import { ParsedIamMemberType } from '../../utils/iam';
+
+export function toIamRoleEntityKey(params: {
+  projectId: string;
+  roleName: string;
+}): string {
+  // Managed roles do not contain the project ID. We just prefix all role
+  // entities with the project ID for safety.
+  return `project:${params.projectId}-role:${params.roleName}`;
```
aiwilliams, line 28: Are managed roles copied into a project, so that every managed role is project-unique? To put it another way, would we not see a single role entity for a managed role, that is `ASSIGNED` to many users across many projects?

src/steps/service-usage/client.ts
```diff
@@ -50,6 +50,7 @@ export class ServiceUsageClient extends Client {
 
     await this.iterateEnabledServices(async (data) => {
       enabledServices.push(data);
+      return Promise.resolve();
```
aiwilliams, line 53: I think `iterateEnabledServices` should accept either an `async` function or a non-`async` function, so that you don't have to do this here. Did the linter catch the need for this?

src/steps/iam/__snapshots__/converters.test.ts.snap
```diff
@@ -152,8 +152,8 @@ Object {
 exports[`#createIamServiceAccountHasKeyRelationship should convert to relationship 1`] = `
 Object {
   "_class": "HAS",
-  "_fromEntityKey": "google_cloud_iam_service_account_1234567890",
-  "_key": "google_cloud_iam_service_account_1234567890|has|google_cloud_iam_service_account_key_projects/j1-gc-integration-dev/serviceAccounts/j1-gc-integration-dev-sa-tf@j1-gc-integration-dev.iam.gserviceaccount.com/keys/12345",
+  "_fromEntityKey": "project:j1-gc-integration-dev-user:j1-gc-integration-dev-sa@j1-gc-integration-dev.iam.gserviceaccount.com",
+  "_key": "project:j1-gc-integration-dev-user:j1-gc-integration-dev-sa@j1-gc-integration-dev.iam.gserviceaccount.com|has|google_cloud_iam_service_account_key_projects/j1-gc-integration-dev/serviceAccounts/j1-gc-integration-dev-sa-tf@j1-gc-integration-dev.iam.gserviceaccount.com/keys/12345",
   "_toEntityKey": "google_cloud_iam_service_account_key_projects/j1-gc-integration-dev/serviceAccounts/j1-gc-integration-dev-sa-tf@j1-gc-integration-dev.iam.gserviceaccount.com/keys/12345",
```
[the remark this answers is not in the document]
austinkelleher: Yes. I plan to do this in a subsequent PR. I didn't want to flood this PR with unnecessary changes. Thanks for raising it though!

src/steps/iam/__snapshots__/index.test.ts.snap
```diff
@@ -7,7 +7,7 @@ Object {
       "_class": Array [
         "AccessRole",
       ],
-      "_key": "google_cloud_iam_role_projects/j1-gc-integration-dev/roles/545240943112customrole",
+      "_key": "project:j1-gc-integration-dev-role:projects/j1-gc-integration-dev/roles/545240943112customrole",
```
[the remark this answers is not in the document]
austinkelleher: Yes, I completely agree. I did think about this. The reason I did it this way is because managed roles do not contain the project id.

Managed: `roles/iam.securityReviewer`
Non-managed: `projects/j1-gc-integration-dev/roles/abc123`

The more I think about it though, it doesn't really matter. Dropping the project ID from the prefix would still yield uniqueness scoped to the integration instance itself despite managed roles not being globally unique. I will remove the prefix.

src/steps/service-usage/client.ts
austinkelleher: Yes. The linter caught this because of the `require-await` rule: https://github.com/JupiterOne/sdk/blob/master/packages/integration-sdk-dev-tools/config/eslint.json#L32

`require-await` prevents functions from being declared that do not actually either return a promise or reference `await`. After that rule was enabled in the SDK, I found a lot of functions that didn't need to be async at all.

The callback in this function does not have to be async in this case. I will change it to support both because the caller could benefit from an async function, but it's not needed for this specific case.

src/steps/resource-manager/index.test.ts
austinkelleher: Agreed! Starting to see more ways that we can improve these test helpers.

austinkelleher:
@aiwilliams Thanks for all of the fantastic feedback. I think I've addressed everything.

src/steps/iam/__snapshots__/converters.test.ts.snap
```diff
@@ -191,3 +191,32 @@ Object {
...
+    "User",
+  ],
+  "_key": "test.user@example.com",
+  "_rawData": Array [
+    Object {
+      "name": "default",
+      "rawData": Object {
+        "deleted": false,
+        "identifier": "test.user@example.com",
+        "projectId": "abc123",
```
aiwilliams, line 207: I think this shouldn't be bound to a project, right?

aiwilliams: To clarify, I think the service account User entities should have this, but not the regular users.

src/steps/resource-manager/__snapshots__/index.test.ts.snap
```diff
@@ -0,0 +1,818 @@
...
+      "createdOn": undefined,
+      "custom": false,
+      "deleted": false,
+      "description": undefined,
+      "displayName": "roles/iam.securityReviewer",
+      "etag": undefined,
+      "name": "roles/iam.securityReviewer",
+      "permissions": undefined,
+      "stage": undefined,
+    },
```
aiwilliams, line 535: I think these should be target entities of mapped relationships, or else each project that refers to them will create their own entity representing this managed role.

austinkelleher:
Thanks for collecting that info @aiwilliams! I opened up a separate PR addressing the `ASSIGNED` feedback you described: https://github.com/JupiterOne/graph-google-cloud/pull/12 (branches off of this PR because I had a bunch of other changes)

Issue #11, opened 2020-08-11
Update all existing entity keys to not include the `_type` prefix


austinkelleher:
See: https://github.com/JupiterOne/sdk/issues/274

PR #12, opened 2020-08-11
Update all existing entity keys to not include the `_type` prefix

austinkelleher:
Also include `projectId` in the user ASSIGNED role relationship.

#11


src/steps/storage/__snapshots__/converters.test.ts.snap
```diff
@@ -5,7 +5,7 @@ Object {
   "_class": Array [
     "DataStore",
   ],
-  "_key": "google_cloud_storage_bucket_customer-managed-encryption-key-bucket-1234",
+  "_key": "bucket:customer-managed-encryption-key-bucket-1234",
```
aiwilliams, line 8: 💯 🔥

PR #19, opened 2020-08-16
Fixes #18 - Ingest compute disks and compute instances

### LINKED ISSUE #18: Ingest Google Compute disks and Google Compute instances
`google_compute_disk`
`google_compute_instance`

Relationship: `google_compute_instance_uses_disk`

austinkelleher:
#18 

NOTE: Compute instances and disks are specific to zones in regions. This PR also introduces the ability to iterate across all regions and each zone in the region.


src/steps/compute/constants.ts
```diff
@@ -0,0 +1,10 @@
+export const STEP_COMPUTE_INSTANCES = 'fetch-compute-instances';
+export const STEP_COMPUTE_DISKS = 'fetch-compute-disks';
+
+export const ENTITY_CLASS_COMPUTE_DISK = 'DataStore';
```
erkangz, line 4: this should be `['DataStore', 'Disk']`

austinkelleher: @erkangz Disk is not a supported class. Did you mean that we should add a `Disk` class or did you mean a different class?

erkangz: let's definitely add `Disk` to the data model if not there. AWS EBS volumes already use that class

PR #25, opened 2020-09-02
Build relationship between compute instance and service account

austinkelleher:
`google_compute_instance` **TRUSTS** `google_iam_service_account`


CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
+- Build relationship between `google_compute_disk` and
```
aiwilliams, line 13:
```suggestion
- Build relationship between `google_compute_instance` and
```

src/steps/compute/index.ts
```diff
@@ -119,8 +162,14 @@ export const computeSteps: IntegrationStep<IntegrationConfig>[] = [
         sourceType: ENTITY_TYPE_COMPUTE_INSTANCE,
         targetType: ENTITY_TYPE_COMPUTE_DISK,
       },
+      {
+        _class: RelationshipClass.USES,
+        _type: 'google_compute_instance_trusts_iam_service_account',
+        sourceType: ENTITY_TYPE_COMPUTE_INSTANCE,
+        targetType: ENTITY_TYPE_COMPUTE_DISK,
+      },
```
aiwilliams, line 170: I think this isn't quite right.

austinkelleher: Whoa whoops. Thanks for pointing that out.

PR #26, opened 2020-09-28
Fix DUPLICATE_KEY_ERR in fetchResourceManagerIamPolicy

ndowmon:
The previous code always created a `User` entity in this step, not taking into account that the previous `PolicyMemberBinding`s could target the same user. Fixed by always checking if a user exists before creating the relationship.

Once I get an approval I'll bump the version (minor to `3.1.0`, instead of patch to `3.0.1`, because there are other unreleased changes noted in the `CHANGELOG`).

### DIFF
--- a/src/steps/resource-manager/index.ts
+++ b/src/steps/resource-manager/index.ts
@@ -53,14 +53,21 @@ async function maybeFindOrCreateIamUserEntity({
     // will just ignore it.
     userEntity = await jobState.findEntity(parsedIdentifier);
   } else {
-    userEntity = await jobState.addEntity(
-      createIamUserEntity({
-        type: parsedMemberType,
-        identifier: parsedIdentifier,
-        uniqueid: parsedMember.uniqueid,
-        deleted: parsedMember.deleted,
-      }),
-    );
+    // We always want to find or create relationships to user entities. Since
+    // the same user can have multiple roles, there is a possibility that a
+    // user is created earlier in this step. In order to avoid duplicate entity
+    // keys, we need to look for the user before creating anything.
+    userEntity = await jobState.findEntity(parsedIdentifier);
+    if (!userEntity) {
+      userEntity = await jobState.addEntity(
+        createIamUserEntity({
+          type: parsedMemberType,
+          identifier: parsedIdentifier,
+          uniqueid: parsedMember.uniqueid,
+          deleted: parsedMember.deleted,
+        }),
+      );
+    }
   }
 
   return userEntity;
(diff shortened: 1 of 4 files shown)


src/steps/resource-manager/index.ts
```diff
@@ -53,14 +53,21 @@ async function maybeFindOrCreateIamUserEntity({
...
-        identifier: parsedIdentifier,
-        uniqueid: parsedMember.uniqueid,
-        deleted: parsedMember.deleted,
-      }),
-    );
+    // We always want to find or create relationships to user entities. Since
+    // the same user can have multiple roles, there is a possibility that a
+    // user is created earlier in this step. In order to avoid duplicate entity
+    // keys, we need to look for the user before creating anything.
+    userEntity = await jobState.findEntity(parsedIdentifier);
```
austinkelleher, line 60: Looks like the above condition also performs the same `findEntity` call. Can we simply this entire function to something like:

```ts
async function maybeFindOrCreateIamUserEntity({
  jobState,
  projectId,
  member,
}: {
  jobState: JobState;
  projectId: string;
  member: string;
}): Promise<Entity | null> {
  const parsedMember = parseIamMember(member);
  const { identifier: parsedIdentifier, type: parsedMemberType } = parsedMember;

  if (!parsedIdentifier || parsedMemberType === 'domain') {
    // This is a member that we do not want to create. For example: `allAuthenticatedUsers`
    // or `domain:{domain}`.
    return null;
  }

  // We always want to find or create relationships to user entities. Since
  // the same user can have multiple roles, there is a possibility that a
  // user is created earlier in this step. In order to avoid duplicate entity
  // keys, we need to look for the user before creating anything.
  const userEntity = await jobState.findEntity(parsedIdentifier);

  return userEntity || (
    await jobState.addEntity(
      createIamUserEntity({
        type: parsedMemberType,
        identifier: parsedIdentifier,
        uniqueid: parsedMember.uniqueid,
        deleted: parsedMember.deleted,
      }),
    )
  )
}
```

austinkelleher (approved):

src/steps/resource-manager/index.ts
ndowmon: Thanks for noticing @austinkelleher. I had set that up originally, but it exposes a number of new `serviceAccount` entities and subsequent relationships. It depends on how much you care about this comment:

```typescript
  if (parsedMember.type === 'serviceAccount') {
    // We already created service accounts in another step. If this service
    // account does not exist, it's possible that it was created in between
    // steps. The caller handles the case that not entity is returned, so we
    // will just ignore it.
    userEntity = await jobState.findEntity(parsedIdentifier);
```

austinkelleher (approved):

PR #30, opened 2020-10-03
Initial network ingestion

austinkelleher:
Screen Shot 2020-10-03 at 3 01 04 PM


src/utils/firewall.ts
```diff
@@ -0,0 +1,137 @@
...
+        ipProtocol: protocol,
+      };
+
+      if (isInternet(ipRange)) {
+        await callback({
+          targetFilterKeys: [['_key']],
+          targetEntity: INTERNET,
+          properties: relationshipProperties,
+        });
+      } else if (isHost(ipRange)) {
```
erkangz, line 125:
```suggestion
      } else if (isHost(ipRange)) {
        if (isPublicIp(ipRange)) {
          await callback({
            targetFilterKeys: [['_class', 'publicIpAddress']],
            targetEntity: {
              _class: 'Host',
              publicIpAddress: ipRange.split('/')[0],
            },
            properties: relationshipProperties,
          });
        }
        else {
          await callback({
            targetFilterKeys: [['_class', 'privateIpAddress']],
            targetEntity: {
              _class: 'Host',
              privateIpAddress: ipRange.split('/')[0],
            },
            properties: relationshipProperties,
          });
        }
      } else {
        await callback({
          targetFilterKeys: [['_class', 'CIDR']],
          targetEntity: {
            _class: 'Network',
            CIDR: ipRange,
            netmask: ipRange.split('/')[1],
          },
          properties: relationshipProperties,
        });
      }
```

erkangz (changes requested):

terraform/compute_vms.tf
```diff
@@ -31,3 +31,88 @@ resource "google_compute_instance" "testvm_not_accesible_internet" {
...
+resource "google_compute_firewall" "deny_ssh" {
+  name    = "${local.public_compute_app_name}-fw-deny-ssh"
+  network = google_compute_network.public_compute_vpc.name
+
+  deny {
+    protocol = "tcp"
+    ports    = ["22"]
+  }
+
+  target_tags = ["https"]
```
erkangz, line 79: should this be `"ssh"`?

erkangz (changes requested):

src/steps/compute/constants.ts
```diff
@@ -1,10 +1,37 @@
+// Steps
 export const STEP_COMPUTE_INSTANCES = 'fetch-compute-instances';
 export const STEP_COMPUTE_DISKS = 'fetch-compute-disks';
+export const STEP_COMPUTE_NETWORKS = 'fetch-compute-networks';
+export const STEP_COMPUTE_SUBNETWORKS = 'fetch-compute-subnetworks';
+export const STEP_COMPUTE_FIREWALLS = 'fetch-compute-firewalls';
 
+// Entities
```
ndowmon, line 8: Not necessary, but it would have been nice to have updated this `constants.ts` file to match the newer style of entity/relationship definitions (using `StepEntityMetadata` and `StepRelationshipMetadata`)

ndowmon, line 8: Not necessary, but it would have been nice to have updated this `constants.ts` file to match the newer style of entity/relationship definitions (using `StepEntityMetadata` and `StepRelationshipMetadata`)

src/steps/compute/converters.ts
```diff
@@ -185,3 +198,253 @@ export function createComputeInstanceTrustsServiceAccountRelationship(params: {
...
+        egress: data.direction === 'EGRESS',
+        logConfigEnabled: data.logConfig?.enable,
+        logConfigMetadata: data.logConfig?.metadata,
+        disabled: data.disabled,
+        destinationRanges: data.destinationRanges,
+        /**
+         * If source service accounts are specified, the firewall rules apply
+         * only to traffic originating from an instance with a service account
+         * in this list. Source service accounts cannot be used to control
+         * traffic to an instance&#39;s external IP address because service
```
ndowmon, line 256:
```suggestion
         * traffic to an instance's external IP address because service
```

src/steps/compute/index.ts
```diff
@@ -98,6 +124,32 @@ async function iterateComputeInstanceServiceAccounts(params: {
...
+  jobState: JobState;
+  computeInstanceEntity: Entity;
+  computeInstance: compute_v1.Schema$Instance;
+}) {
+  const { jobState, computeInstanceEntity, computeInstance } = params;
+
+  for (const networkInterface of computeInstance.networkInterfaces || []) {
+    const subnetwork = networkInterface.subnetwork as string;
+
+    const subnetEntity = await jobState.findEntity(subnetwork);
```
ndowmon, line 137: I noticed that this addition added two steps to the `dependsOn` of `fetchComputeInstances`-  `STEP_COMPUTE_NETWORKS` _and_ `STEP_COMPUTE_SUBNETWORKS`.

Can a `google_compute_network` have a `google_compute_instance`, or can only a `google_compute_subnetwork` have a `google_compute_instance`?

terraform/output.tf
```diff
@@ -0,0 +1,7 @@
+output "public_vm_external_ip" {
```
ndowmon, line 1: What's the point of these outputs? Maybe I missed something but I didn't see them referenced. Also, does it make sense to put these in an `output.tf` file, or to output them from `compute_vms.tf` where they are generated?

ndowmon (commented):
Looks good @austinkelleher. I left a few notes.

terraform/output.tf
austinkelleher: I mainly did this so that I would get a nice log output at the end of my terraform apply with the public IP so I could test it out.

src/steps/compute/constants.ts
austinkelleher: Agreed. Going to do this in bulk separately.

erkangz (approved):

PR #31, opened 2020-10-04
Support for ingesting KSM crypto key rings and crypto keys

austinkelleher:
NOTE: This branches off of `initial-networking`

### DIFF
--- a/docs/jupiterone.md
+++ b/docs/jupiterone.md
@@ -78,6 +78,8 @@ The following entities are created:
 | IAM Service Account     | `google_iam_service_account`     | `User`              |
 | IAM Service Account Key | `google_iam_service_account_key` | `AccessKey`         |
 | IAM User                | `google_user`                    | `User`              |
+| KMS Crypto Key          | `google_kms_crypto_key`          | `Key`, `CryptoKey`  |
+| KMS Key Ring            | `google_kms_key_ring`            | `Vault`             |
 
 ### Relationships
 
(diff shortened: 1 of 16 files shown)


src/steps/kms/constants.ts
```diff
@@ -0,0 +1,10 @@
+export const STEP_CLOUD_KMS_KEY_RINGS = 'fetch-kms-key-rings';
```
ndowmon, line 1: Same question as I posed in your network resources [PR](https://github.com/JupiterOne/graph-google-cloud/pull/30#discussion_r499273243) - just wondering if it'd be nice to start moving in the direction of 

```typescript
const Entities = {
  KMS_KEY_RING: {
    resourceName: 'KMS Key Ring',
    _type: 'google_kms_key_ring',
    _class: 'Vault',
  },
  KMS_CRYPTO_KEY: {
    resourceName: 'KMS Crypto Key',
    _type: 'google_kms_crypto_key',
    _class: ['Key', 'CryptoKey'],
  },
}
```

Your call. Obviously this would some inconsistencies with the rest of the services you've already implemented.

ndowmon (commented):
Just a few small notes.

docs/jupiterone.md
```diff
@@ -94,6 +96,7 @@ The following relationships are created/mapped:
 | `google_compute_subnetwork`  | **HAS**               | `google_compute_instance`        |
 | `google_iam_service_account` | **ASSIGNED**          | `google_iam_role`                |
 | `google_iam_service_account` | **HAS**               | `google_iam_service_account_key` |
+| `google_kms_key_ring`        | **HAS**               | `google_kms_crypto_key`          |
```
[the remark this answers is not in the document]
austinkelleher: There is no relationship between service account and crypto key/ring. Service account keys are different. We have a `google_iam_service_account_key` for that. We should consider creating the relationship with the account, but I'll tackle this separately.

src/steps/kms/constants.ts
austinkelleher: Thanks. I plan to do this in bulk separately.

ndowmon (approved):
Thanks for responding to my comments. Looks good.

PR #34, opened 2020-10-08
Add public property to google_storage_bucket

austinkelleher:
We should also ingest the policy and build a relationship between the policy and the bucket, but this will help us answer:

```
# CIS 5.1 Ensure that Cloud Storage bucket is not anonymously or publicly accessible
find google_compute_bucket with public=true
```

### DIFF
--- a/src/steps/storage/converters.ts
+++ b/src/steps/storage/converters.ts
@@ -22,7 +22,7 @@ export function createCloudStorageBucketEntity({
 }: {
   data: storage_v1.Schema$Bucket;
   projectId: string;
-  isPublic?: boolean;
+  isPublic: boolean;
 }) {
   return createIntegrationEntity({
     entityData: {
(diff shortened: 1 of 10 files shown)


src/steps/storage/converters.ts
```diff
@@ -15,10 +15,15 @@ function getCloudStorageBucketWebLink(
...
-  projectId: string,
-) {
+export function createCloudStorageBucketEntity({
+  data,
+  projectId,
+  isPublic,
+}: {
+  data: storage_v1.Schema$Bucket;
+  projectId: string;
+  isPublic?: boolean;
```
ndowmon, line 25: Should this really be optional?

ndowmon (approved):
Looks good to me, just posed one Q. Approving anyway.

src/steps/storage/converters.ts
austinkelleher: Nope this was left on accident. Originally it needed to be optional, but I changed some stuff...Will fix. Thank you!

ndowmon (approved):

PR #40, opened 2020-10-22
Fix typo

ndowmon:
Aesthetic only, no functional change.

Fix typo introduced in #39

### DIFF
--- a/src/steps/resource-manager/index.ts
+++ b/src/steps/resource-manager/index.ts
@@ -158,11 +158,11 @@ export async function fetchResourceManagerProject(
     const message =
       'Could not fetch project from Cloud Resource Manager API. Ensure the API is enabled (see https://github.com/JupiterOne/graph-google-cloud/blob/master/docs/development.md#enabling-google-cloud-services)';
     if (err.code === 403) {
-      logger.warn({ err }, message),
-        logger.publishEvent({
-          name: 'FETCH_PROJECT_ERROR',
-          description: err.message || `403: ${message}`,
-        });
+      logger.warn({ err }, message);
+      logger.publishEvent({
+        name: 'FETCH_PROJECT_ERROR',
+        description: err.message || `403: ${message}`,
+      });
     } else {
       logger.publishErrorEvent({
         name: 'FETCH_PROJECT_ERROR',


src/steps/resource-manager/index.ts
```diff
@@ -158,11 +158,11 @@ export async function fetchResourceManagerProject(
     const message =
       'Could not fetch project from Cloud Resource Manager API. Ensure the API is enabled (see https://github.com/JupiterOne/graph-google-cloud/blob/master/docs/development.md#enabling-google-cloud-services)';
     if (err.code === 403) {
-      logger.warn({ err }, message),
-        logger.publishEvent({
-          name: 'FETCH_PROJECT_ERROR',
-          description: err.message || `403: ${message}`,
-        });
+      logger.warn({ err }, message);
```
austinkelleher, line 161: I'm wondering if we really want to even log this as a warn as it's strictly a lack of configuration on the user's side, right?

austinkelleher: Might be worth logging as "info" though. It might be worth having a strategy to raise this type of thing internally so that our Solutions Engineering team can communicate with customers when things are not configured properly. Not sure exactly...

src/steps/resource-manager/index.ts
```diff
@@ -158,11 +158,11 @@ export async function fetchResourceManagerProject(
...
+      logger.publishEvent({
+        name: 'FETCH_PROJECT_ERROR',
+        description: err.message || `403: ${message}`,
+      });
     } else {
       logger.publishErrorEvent({
```
austinkelleher, line 167: Is there a case where this could be a legit error that our team would be interested in looking at? I don't think `publishErrorEvent` currently actually logs anything on our end. Would be worth double checking.

ndowmon: It does log an error on our end before emitting to the user.

src/steps/resource-manager/index.ts
ndowmon: Maybe it should be an `info` level. For the record, we currently log validation failures at `warn` level: https://github.com/JupiterOne/sdk/blob/master/packages/integration-sdk-runtime/src/logger/index.ts#L307

I do think we should log when this happens, though.

ndowmon: Especially considering our config validation will probably not be 100% accurate, so this handling _could_ actually have nothing to do with their configuration, and I think hiding it from ourselves would cripple us in that event.

Also, in this _particular_ step, I think it's good to have a log telling us if they have a real `Account` entity or just a synthetic one created with `projectId`.

aiwilliams: What I'd love to have is metrics per endpoint, per account, per response code and avoid logging, but this is a very expensive feature in CloudWatch metrics. 😬 Also, would be great to have complete logs for a window of time so that customers can see what's going on. We could retire the detailed logs after a day or two or after 3 successful runs. Anyway, more food for thought, to the end that we have operable software, shorten MTTR, and enable customers to support themselves as much as possible!

aiwilliams (approved):

Issue #38, opened 2020-10-22
Some integrations throw on fetching `project`


ndowmon:
A recently introduced integration step fetches the `project` details to create a root `Account` entity in the integration. However, users may not have permissions to fetch the `project` entity, if they have not enabled `cloudresourcemanager` APIs or for some other permissions reason.

PR #46, opened 2020-12-10
Improve experience of '403: Caller does not have permission' error

ndowmon:
Previously, if `err.message` existed and the error was a 403, we would _only_ display `Caller does not have permission` in the job log. Now, we will display the full remediation message:

`Could not fetch project from Cloud Resource Manager API. Ensure the API is enabled (see 'https://github.com/JupiterOne/graph-google-cloud/blob/master/docs/development.md#enabling-google-cloud-services) {code:403,message:'Caller does not have permission'}`

Related to #38

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,15 +8,19 @@ and this project adheres to
 
 ## [Unreleased]
 
+## 0.10.0 - 2020-12-11
+
+### Changed
+
 - Improve job log messaging when `/v1/projects/{projectId}` request responds
   with 403:FORBIDDEN. Previously, the job log simply stated "The caller does not
   have permission".
+
+### Added
+
 - Add `roles/iam.roleViewer` as a required role in developer documentation. This
   role includes the `resourcemanager.projects.get` permission, which is required
   to access the `/v1/projects/{projectId}` endpoint.
-
-## 0.10.0 - 2020-11-11
-
 - Improve JupiterOne Google Cloud organization script to walk all folders in an
   organization.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ and this project adheres to
 
 ## [Unreleased]
 
+- Improve job log messaging when `/v1/projects/{projectId}` request responds
+  with 403:FORBIDDEN. Previously, the job log simply stated "The caller does not
+  have permission".
+- Add `roles/iam.roleViewer` as a required role in developer documentation. This
+  role includes the `resourcemanager.projects.get` permission, which is required
+  to access the `/v1/projects/{projectId}` endpoint.
+
 ## 0.10.0 - 2020-11-11
 
 - Improve JupiterOne Google Cloud organization script to walk all folders in an
--- a/src/steps/resource-manager/index.ts
+++ b/src/steps/resource-manager/index.ts
@@ -162,12 +162,12 @@ export async function fetchResourceManagerProject(
     if (err.code === 403) {
       logger.warn({ err }, message);
       logger.publishEvent({
-        name: 'GET_PROJECT_FORBIDDEN_ERROR',
+        name: 'get_project_forbidden_error',
         description: message,
       });
     } else {
       logger.publishErrorEvent({
-        name: 'GET_PROJECT_ERROR',
+        name: 'get_project_error',
         message,
         err,
       });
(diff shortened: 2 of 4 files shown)


src/steps/resource-manager/index.ts
```diff
@@ -155,16 +155,18 @@ export async function fetchResourceManagerProject(
     // This step _always_ executes because it creates a root `Account` entity for the integration instance.
     // However, users can only fetch the project details if cloudresourcemanager API is enabled.
     const message =
-      'Could not fetch project from Cloud Resource Manager API. Ensure the API is enabled (see https://github.com/JupiterOne/graph-google-cloud/blob/master/docs/development.md#enabling-google-cloud-services)';
+      'Could not fetch project from Cloud Resource Manager API. Ensure the API is enabled (see ' +
+      'https://github.com/JupiterOne/graph-google-cloud/blob/master/docs/development.md#enabling-google-cloud-services) ' +
+      `${JSON.stringify({ code: err.code, message: err.message })}`;
```
aiwilliams, line 160: Not a big deal at this time, but for consistency sake and so that others are aware and we can work toward a solution over time, I'll point out the formatting of [messages from the SDK](https://github.com/JupiterOne/sdk/blob/8ff1018a6abceb131bf0c398aaa49ac43d4cc5e3/packages/integration-sdk-runtime/src/execution/__tests__/dependencyGraph.test.ts#L649). I think we'll need to work toward an API where we can provide a message and "parenthetical" properties (as well as `error: true`!) that are stored with the event and that are consistently presented by the job log UI.

src/steps/resource-manager/index.ts
```diff
@@ -155,16 +155,18 @@ export async function fetchResourceManagerProject(
...
     if (err.code === 403) {
       logger.warn({ err }, message);
       logger.publishEvent({
-        name: 'FETCH_PROJECT_ERROR',
-        description: err.message || `403: ${message}`,
+        name: 'GET_PROJECT_FORBIDDEN_ERROR',
```
aiwilliams, line 164: I'd like to see all job event names match the pattern of lower snake case. Until this project, none others were using this form.

aiwilliams: Also, note that anything that matches `/error/i` will mark the job as `errorsOccurred`. `*_failure` does not. 😢  This is why we need an API where we send the event as `error: true` (probably would be good to have `warn: true` as well, or `level: 'info' | 'warn' | 'error'`).

aiwilliams (changes requested):

src/steps/resource-manager/index.ts
ndowmon: I have heard you mention that about `error` vs `failure`. Seems like regexing `/error/i` out of a job log is not the most foolproof way to identify if an error occurred...

```bash
2020-12-10-17-56-30UTC [step-success] Step "fetch-system-errors" completed successfully"
```

ndowmon: To be clear, your requested change is to downcase `get_project_forbidden_error` and `get_project_error`?

aiwilliams: > I have heard you mention that about `error` vs `failure`. Seems like regexing `/error/i` out of a job log is not the most foolproof way to identify if an error occurred...

Definitely not foolproof! I think that code is nearly 3 years old 😳

aiwilliams: > To be clear, your requested change is to downcase `get_project_forbidden_error` and `get_project_error`?

I didn't think beyond the case at the time of my comment, but now I wonder if we should call these `auth_error` events? It could be worthwhile to search for calls to `publishEvent` for a survey of names. The key point here is, I think we want to end up with a common set of event names. What do you thing?

ndowmon: I agree. And I don't really like the names I had initially given, they are too specific.

CHANGELOG.md
```diff
@@ -8,6 +8,13 @@ and this project adheres to
 
 ## [Unreleased]
 
+- Improve job log messaging when `/v1/projects/{projectId}` request responds
+  with 403:FORBIDDEN. Previously, the job log simply stated "The caller does not
+  have permission".
+- Add `roles/iam.roleViewer` as a required role in developer documentation. This
+  role includes the `resourcemanager.projects.get` permission, which is required
+  to access the `/v1/projects/{projectId}` endpoint.
```
aiwilliams, line 16: 🔥

aiwilliams (approved):

aiwilliams (approved):

PR #50, opened 2020-12-22
Add aim.roleViewer requirement to docs

a-u-h-g:
New role required for GCP integration to ingest without error


ndowmon (approved):
Thanks for fixing this!

PR #51, opened 2021-01-05
AB#2089 handle errors in step handler functions

mknoedel:
Add error handling and handle billing configuration errors outside of Sentury


src/utils/withErrorHandling.test.ts
```diff
@@ -0,0 +1,35 @@
...
+import { IntegrationStepContext } from '../types';
+import { withErrorHandling } from './withErrorHandling';
+const { promisify } = require('util');
+
+test('should throw an IntegrationValidationError when a billing message is error is thrown', async () => {
+  const throwerFunction = promisify((context: IntegrationStepContext) => {
+    throw new Error(
+      'This API method requires billing to be enabled. Please enable billing on project #545240943112 by visiting https://console.developers.google.com/billing/enable?project=545240943112 then retry. If you enabled billing for this project recently, wait a few minutes for the action to propagate to our systems and retry.',
+    );
+  });
```
aiwilliams, line 14: This can be achieved with Jest:
```suggestion
  const executionHandler = jest.fn().mockRejectedValue(
    new Error('This API method requires billing to be enabled. Please enable billing on project #545240943112 by visiting https://console.developers.google.com/billing/enable?project=545240943112 then retry. If you enabled billing for this project recently, wait a few minutes for the action to propagate to our systems and retry.'));
```

src/utils/withErrorHandling.test.ts
```diff
@@ -0,0 +1,35 @@
...
+  const handledThrowerFunction = withErrorHandling(throwerFunction);
+  const fakeContext = {} as IntegrationStepContext;
+  try {
+    await handledThrowerFunction(fakeContext);
+  } catch (error) {
+    expect(error).toBeInstanceOf(IntegrationValidationError);
+  }
```
aiwilliams, line 21: With Jest, this can be:
```suggestion
  await expect(handledThrowerFunction(fakeContext))
    .rejects.toThrow(IntegrationValidationError);
```

aiwilliams (commented):
Needs a `CHANGELOG.md` update.

austinkelleher (commented):
Besides the testing comments that Adam mentioned, this looks great! Thanks for the fixes.

src/utils/withErrorHandling.ts
```diff
@@ -0,0 +1,28 @@
+import {
+  IntegrationError,
+  IntegrationValidationError,
+} from '@jupiterone/integration-sdk-core';
+import { IntegrationStepContext } from '../types';
+
+/**
+ * Handles errors for step handler functions
+ */
+export function withErrorHandling(
```
ndowmon, line 10: This feels like a pretty broad stroke to take, wrapping each execution handler with `withErrorHandling()`. I imagine that these errors matching `/requires\sbilling/` can be isolated to the API calls... why not handle this case in `src/google-cloud/client.ts` instead?

mknoedel: From the logs, it looks like every step is throwing this error which means we are seeing it 6 times on every integration run. I believe this is a standard pattern for google apis so we likely want to handle it for every api call.

also from what I've seen, we aren't handling many errors in this project which means whenever there is an error thrown anywhere, it gets posted to the job log as a scary "Unexpected Error has occurred" and also gets sent to Sentry. I think it makes sense to have each integration handle their own errors and this gives us a common place to do so.

src/utils/withErrorHandling.test.ts
aiwilliams: Did this not work out after all?

src/utils/withErrorHandling.ts
aiwilliams: I understand where you're coming from @ndowmon, having had a similar feeling when I reviewed the code. I think it could be fine to do it this way (perhaps we need an `onStepFailure` callback for invocation configs so we don't have to wrap the `executionHandler` everywhere), though it is worth documenting the reason that the API client doesn't handle this, since the pattern we've been pursuing is one where the clients know the nuances of the API responses. There is an [`IntegrationProviderAuthorizationError`](https://github.com/JupiterOne/sdk/blob/8a3c4bea6b65fc8819a0208018443d69a6628796/packages/integration-sdk-core/src/errors.ts#L224) that can be thrown from the client and it will be shown to the user. It may be that the client in this project needs to be refactored to wrap the calls.

austinkelleher (dismissed):

src/utils/withErrorHandling.ts
aiwilliams: Another thought that came to me... The [`IntegrationValidationError`](https://github.com/JupiterOne/sdk/blob/8a3c4bea6b65fc8819a0208018443d69a6628796/packages/integration-sdk-core/src/errors.ts#L159) is documented and intended to be thrown from the `validateInvocation` function, so it may lead to confusion to throw it from not that function.

Seems this is an authorization error, not an authentication one and not a validation one.

Sorry for the late comment, but figured it would be helpful for consistency and clarity 😁

ndowmon: @aiwilliams: Michael and I had a call this morning about the design here. I think he's working on some of these changes

mknoedel: Yeah I agree and switched it to an authorization error. Good catch @aiwilliams thanks!

src/utils/withErrorHandling.test.ts
mknoedel: That does work thanks!

src/google-cloud/client.test.ts
```diff
@@ -66,3 +70,58 @@ describe('#getAuthenticatedServiceClient', () => {
     expect(mockGetAccessToken).toHaveBeenCalledWith();
   });
 });
+
+describe('withErrorHandling', () => {
+  test('should throw an IntegrationValidationError when a billing message is error is thrown', async () => {
```
ndowmon, line 75: Can you update the name of this test? It now checks for `IntegrationProviderAuthorizationError`, not `IntegrationValidationError`.

src/google-cloud/client.test.ts
```diff
@@ -66,3 +70,58 @@ describe('#getAuthenticatedServiceClient', () => {
...
+      );
+    const handledFunction = withErrorHandling(executionHandler);
+    try {
+      await handledFunction();
+    } catch (error) {
+      expect(error).toBeInstanceOf(IntegrationProviderAuthorizationError);
+    }
+  });
+
+  test('should throw an IntegrationValidationError when another billing message is error is thrown', async () => {
```
ndowmon, line 91: You'll want to update the name of this test as well.

src/google-cloud/client.test.ts
```diff
@@ -66,3 +70,58 @@ describe('#getAuthenticatedServiceClient', () => {
...
+      );
+    const handledFunction = withErrorHandling(executionHandler);
+    try {
+      await handledFunction();
+    } catch (error) {
+      expect(error).toBeInstanceOf(IntegrationProviderAuthorizationError);
+    }
+  });
+
+  test('should throw an IntegrationError on all other unhandled errors', async () => {
```
ndowmon, line 107: Should be `IntegrationProviderAPIError`, not `IntegrationError`

src/google-cloud/client.ts
```diff
@@ -24,14 +28,28 @@ export type PageableGaxiosResponse<T> = GaxiosResponse<
 
 export type GoogleClientAuth = JWT | Compute | UserRefreshClient;
 
+export function withErrorHandling<T extends (...params: any) => any>(fn: T) {
+  return async (...params: any) => {
+    try {
+      return await fn(...params);
+    } catch (error) {
+      if (error.message && error.message.match(/billing/i)) {
+        throw new IntegrationProviderAuthorizationError(error.message);
```
ndowmon, line 37: Check out the signature of `IntegrationProviderAuthorizationError` and `IntegrationProviderApiError` in the sdk: https://github.com/JupiterOne/sdk/blob/master/packages/integration-sdk-core/src/errors.ts#L182

Not sure if you have access to the endpoint at this level. May need to find a way to access that. @austinkelleher are the endpoint paths accessible?

```typescript
export declare class IntegrationProviderAuthorizationError extends IntegrationError {
    constructor(options: {
        cause?: Error;
        endpoint: string;
        status: string | number;
        statusText: string;
        resourceType?: string[];
    });
}

export declare class IntegrationProviderAPIError extends IntegrationError {
    constructor(options: {
        cause?: Error;
        endpoint: string;
        status: string | number;
        statusText: string;
    });
}
```

ndowmon (changes requested):

src/utils/withErrorHandling.test.ts
mknoedel: This did work, just forgot to resolve

aiwilliams (dismissed):
Thanks for the thorough tests 👍

src/google-cloud/client.ts
```diff
@@ -71,3 +76,53 @@ export class Client {
...
+    err = new IntegrationProviderAPIError(errorProps);
+  }
+  if (shouldKeepErrorMessage(error)) {
+    err.message = error.message;
+  }
+  throw err;
+}
+
+function shouldKeepErrorMessage(error: any) {
+  const chillMessageStrings = [
```
ndowmon, line 115: This isn't a big deal but I probably would have called this `disabledBillingMessageStrings` or `errorMessagesToKeep` if we want it to be more generic.

ndowmon (commented):
Looks good. Can you just type the `createErrorProps` function before merging?

src/google-cloud/client.ts
mknoedel: I'll update that

test/recording.ts
```diff
@@ -166,3 +170,24 @@ function redact(entry): void {
...
   }
 }
+
+export async function withRecording(
+  recordingName: string,
+  cb: () => Promise<void>,
+  options?: SetupRecordingInput['options'],
+) {
+  const recording = setupGoogleCloudRecording({
+    directory: __dirname,
```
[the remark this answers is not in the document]
mknoedel: Yeah, that makes sense, I'll add a parameter to allow a user to pass in a directory if they prefer.

ndowmon (approved):

austinkelleher (approved):

PR #64, opened 2021-02-03
Retry concurrent policy modification error

austinkelleher:


scripts/organizationSetup.ts
```diff
@@ -403,6 +404,69 @@ async function createJupiterOneIntegrationInstance(
...
+      await setIamPolicyForProject(googleAccessToken, projectId, newPolicy);
+      logger.info('Successfully updated IAM policy for project...');
+    },
+    {
+      delay: 500,
+      maxAttempts: 5,
+      factor: 1.1,
+      handleError(err, attemptContext) {
+        if (
+          (err.code === 409 || err.code === 'ALREADY_EXISTS') &&
```
austinkelleher, line 455: The value of the `.code` property on the err object was not immediately clear to me from reading the docs. I log the actual full error below and the `err.code` inside of this block, so we should be able to reduce this condition logic later.

aiwilliams (approved):

PR #66, opened 2021-02-03
Make org script more idempotent

austinkelleher:


scripts/organizationSetup.ts
```diff
@@ -27,13 +27,19 @@ const PROJECT_SERVICES_TO_ENABLE: string[] = [
   'cloudresourcemanager.googleapis.com',
   'compute.googleapis.com',
   'cloudkms.googleapis.com',
+  'sqladmin.googleapis.com',
+  'bigquery.googleapis.com',
+  'dns.googleapis.com',
+  'logging.googleapis.com',
+  'monitoring.googleapis.com',
```
austinkelleher, line 34: Bonus changes to add new services we will be supporting

scripts/organizationSetup.ts
```diff
@@ -365,13 +393,59 @@ mutation CreateIntegrationInstance($instance: CreateIntegrationInstanceInput!) {
...
+  integrationInstances(definitionId: $definitionId) {
+    instances {
+      id
+      name
+      accountId
+      pollingInterval
+      integrationDefinitionId
+      description
+      offsiteComplete
+    }
```
aiwilliams, line 407: I guess this will have to be maintained to function properly should new properties be added to the instance records?

scripts/organizationSetupCli.ts
```diff
@@ -74,6 +75,13 @@ cli
       default: false,
     },
   )
+  .option(
+    '--rotate-service-account-keys [rotateServiceAccountKeys]',
+    '(Optional) Creates a new service account key for the JupiterOne service account, deletes the old one, and updates the JupiterOne integration instance',
```
aiwilliams, line 80: This means we'll accept the risk of integrations that are already executing failing, right?

scripts/organizationSetup.ts
```diff
@@ -401,7 +476,218 @@ async function createJupiterOneIntegrationInstance(
...
+      jupiteroneAccountId,
+      jupiteroneApiKey,
+      jupiteroneEnv,
+      integrationInstanceId: existingIntegrationInstance.id,
+      integrationInstanceUpdate: {
+        name: existingIntegrationInstance.name,
+        description: existingIntegrationInstance.description,
+        pollingInterval: existingIntegrationInstance.pollingInterval,
+        config: {
+          config: {
```
aiwilliams, line 667: Does this mean there is a `instance.config.config`, or is this just a confusing API?

aiwilliams (approved):
A few questions, but looks solid!

scripts/organizationSetupCli.ts
austinkelleher: Whoops. I need to update the description. It does _not_ delete old keys.

scripts/organizationSetup.ts
austinkelleher: Thanks for catching that. It didn't manifest well.

scripts/organizationSetup.ts
austinkelleher: Unfortunately the GraphQL type for `UpdateIntegrationInstanceInput` is not exactly the same as an `IntegrationInstance` (only subset of properties), so I'm not exactly sure how we plan to maintain it. Spreading all of the properties is not really an option either because GraphQL will reject if you provide invalid properties to the API...Open to suggestions, but we could iterate separately I think. Thanks for raising.

aiwilliams (approved):

PR #75, opened 2021-02-17
Publish integration job log event when compute.projects.get missing

austinkelleher:
- Update CHANGELOG.md
- Update setup doc supported services

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,40 @@ and this project adheres to
 - Support for `--skip-project-id-regex` in the `jupiterone-organization-setup`
   CLI
 
+- Support for ingesting the following **new** resources
+
+  - Logging
+    - `google_logging_metric`
+    - `google_logging_project_sink`
+    - `google_monitoring_alert_policy`
+  - Networking
+    - `google_compute_health_check`
+    - `google_compute_instance_group`
+    - `google_compute_url_map`
+    - `google_compute_backend_service`
+    - `google_compute_backend_bucket`
+    - `google_compute_target_ssl_proxy`
+    - `google_compute_target_https_proxy`
+    - `google_compute_target_http_proxy`
+    - `google_compute_ssl_policy`
+  - GKE (Google Kubernetes Enginer):
+    - `google_container_cluster`
+
+- New properties added to various existing resources:
+  - Storage
+    - `google_storage_bucket`
+      - `retentionPolicyEnabled`
+      - `retentionPeriod`
+      - `retentionDate`
+  - `google_kms_crypto_key`
+    - `public`
+
+### Fixed
+
+- Log integration job event from the `fetch-compute-project` step when the
+  service account used to execute the integration does not have the
+  `compute.projects.get` permission.
+
 ## 0.12.0 - 2020-02-05
 
 ### Added
--- a/docs/jupiterone.md
+++ b/docs/jupiterone.md
@@ -32,6 +32,12 @@ ingest all of the supported data into JupiterOne:
 | [Cloud Resource Manager](https://console.developers.google.com/apis/library/cloudresourcemanager.googleapis.com) | cloudresourcemanager.googleapis.com |
 | [Cloud Engine](https://console.developers.google.com/apis/library/compute.googleapis.com)                        | compute.googleapis.com              |
 | [Cloud Key Management Service (KMS)](https://console.developers.google.com/apis/library/cloudkms.googleapis.com) | cloudkms.googleapis.com             |
+| [Cloud SQL](https://console.developers.google.com/apis/library/sqladmin.googleapis.com)                          | sqladmin.googleapis.com             |
+| [BigQuery](https://console.developers.google.com/apis/library/bigquery.googleapis.com)                           | bigquery.googleapis.com             |
+| [Cloud DNS](https://console.developers.google.com/apis/library/dns.googleapis.com)                               | dns.googleapis.com                  |
+| [Kubernetes Engine](https://console.developers.google.com/apis/library/container.googleapis.com)                 | container.googleapis.com            |
+| [Cloud Logging](https://console.developers.google.com/apis/library/logging.googleapis.com)                       | logging.googleapis.com              |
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -12,7 +12,7 @@
     "access": "public"
   },
   "scripts": {
-    "start": "j1-integration collect",
+    "start": "LOG_LEVEL=trace j1-integration collect",
```
austinkelleher, line 15: So we can see some of the high-level permission errors when developing locally without actually logging these in CW.

src/steps/compute/index.ts
```diff
@@ -210,10 +211,32 @@ async function iterateComputeInstanceNetworkInterfaces(params: {
...
   const client = new ComputeClient({ config: instance.config });
 
-  const computeProject = await client.fetchComputeProject();
+  let computeProject: compute_v1.Schema$Project;
+
+  try {
+    computeProject = await client.fetchComputeProject();
+  } catch (err) {
+    if (err.code === 403) {
+      logger.trace(
```
austinkelleher, line 223: Purposely `trace` as per message above

docs/jupiterone.md
```diff
@@ -32,6 +32,12 @@ ingest all of the supported data into JupiterOne:
 | [Cloud Resource Manager](https://console.developers.google.com/apis/library/cloudresourcemanager.googleapis.com) | cloudresourcemanager.googleapis.com |
 | [Cloud Engine](https://console.developers.google.com/apis/library/compute.googleapis.com)                        | compute.googleapis.com              |
 | [Cloud Key Management Service (KMS)](https://console.developers.google.com/apis/library/cloudkms.googleapis.com) | cloudkms.googleapis.com             |
+| [Cloud SQL](https://console.developers.google.com/apis/library/sqladmin.googleapis.com)                          | sqladmin.googleapis.com             |
```
austinkelleher, line 35: Bonus changes!

CHANGELOG.md
```diff
@@ -13,6 +13,40 @@ and this project adheres to
 - Support for `--skip-project-id-regex` in the `jupiterone-organization-setup`
   CLI
 
+- Support for ingesting the following **new** resources
```
austinkelleher, line 16: Bonus changes!

src/utils/events.ts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+export function publishMissingPermissionEvent({
+  logger,
+  permission,
+  stepId,
+}: PublishMissingPermissionEventParams) {
+  logger.publishEvent({
+    name: 'missing_permission',
+    description: `"compute.projects.get" is not a required permission to run the Google Cloud integration, but is required for step "${stepId}"`,
```
ndowmon, line 16:
```suggestion
    description: `${permission} is not a required permission to run the Google Cloud integration, but is required for step "${stepId}"`,
```

src/steps/compute/index.ts
```diff
@@ -210,10 +211,27 @@ async function iterateComputeInstanceNetworkInterfaces(params: {
...
   const client = new ComputeClient({ config: instance.config });
 
-  const computeProject = await client.fetchComputeProject();
+  let computeProject: compute_v1.Schema$Project;
+
+  try {
+    computeProject = await client.fetchComputeProject();
+  } catch (err) {
+    if (err.code === 403) {
+      publishMissingPermissionEvent({
```
ndowmon, line 228: I like that you've made a generic `publishMissingPermissionEvent` publisher. Do you think this belongs here, or in the `ComputeClient.fetchComputeProjects()` code?

ndowmon (commented):
I'm requesting a change in `publishMissingPermissionEvent`. I'll approve after that, but I wonder your thoughts about moving this `403` handling code somewhere more generic, either inside `src/steps/compute/client.ts` or inside `src/google-cloud/client.ts`.

In addition, we now have 3 unique places where `403` is handled in this project. I don't think that needs to be rectified before this change goes out, but it's probably worth opening an issue and planning to burn down that tech debt soon.

ndowmon (approved):

ndowmon (approved):

src/steps/compute/index.ts
aiwilliams: Personally, I've been hoping to avoid having the Client objects get references to the logger, and to avoid having them send events to the job log. I would prefer the steps/integration code register for lifecycle events of the clients. This would allow us to drive toward a client toolchain/helper functions over time, and perhaps even offer them as separate NPM packages that are not coupled to our integration SDK codes.

CHANGELOG.md
```diff
@@ -13,6 +13,40 @@ and this project adheres to
...
+    - `google_compute_health_check`
+    - `google_compute_instance_group`
+    - `google_compute_url_map`
+    - `google_compute_backend_service`
+    - `google_compute_backend_bucket`
+    - `google_compute_target_ssl_proxy`
+    - `google_compute_target_https_proxy`
+    - `google_compute_target_http_proxy`
+    - `google_compute_ssl_policy`
+  - GKE (Google Kubernetes Enginer):
```
aiwilliams, line 32:
```suggestion
  - GKE (Google Kubernetes Engine):
```

austinkelleher: Thanks. Will fix this in follow-up PR.

package.json
aiwilliams: I think we should consider using a `debug` level in the integrations, and `trace` in the SDK/libs. This way, we can allow developers of the integrations to emit things without having to see any noise from libraries.

src/utils/events.ts
```diff
@@ -0,0 +1,18 @@
...
+  stepId: string;
+}
+
+export function publishMissingPermissionEvent({
+  logger,
+  permission,
+  stepId,
+}: PublishMissingPermissionEventParams) {
+  logger.publishEvent({
+    name: 'missing_permission',
```
aiwilliams, line 15: I'd like to see a common set of event names. Here I would use `auth` for any kind of authentication or authorization message.

aiwilliams: It's one of the reasons that `publishEvent` is marked as deprecated, but we could undeprecate and just create a known set of event names through an enum.

PR #77, opened 2021-02-17
Add required versions to Terraform

austinkelleher:
@eXtremeX

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,8 @@
 node_modules/
 dist/
 .j1-integration/
-.env
+.env*
 .eslintcache
 tsconfig.tsbuildinfo
 coverage/
+main.remote.tf
--- a/terraform/cloud_storage_buckets.tf
+++ b/terraform/cloud_storage_buckets.tf
@@ -10,60 +10,6 @@ resource "google_kms_crypto_key" "bucket_key" {
   depends_on = [google_kms_key_ring.bucket_key_ring]
 }
 
-resource "google_project_iam_member" "grant_google_storage_service_encrypt_decrypt" {
-  role = "roles/cloudkms.cryptoKeyEncrypterDecrypter"
-  member = "serviceAccount:service-${var.project_id_number}@gs-project-accounts.iam.gserviceaccount.com"
-}
-
-resource "google_storage_bucket" "with_customer_encryption_key" {
-  name = "customer-managed-encryption-key-bucket-${data.google_project.current.number}"
-
-  encryption {
-    default_kms_key_name = google_kms_crypto_key.bucket_key.self_link
-  }
-
-  depends_on = [google_project_iam_member.grant_google_storage_service_encrypt_decrypt]
-}
-
-################################################################################
-# Start public bucket code
-################################################################################
-data "google_iam_policy" "public_storage_object_viewer" {
-  binding {
-    role = "roles/storage.legacyBucketOwner"
-    members = [
-      "projectEditor:j1-gc-integration-dev",
-      "projectOwner:j1-gc-integration-dev"
-    ]
-  }
-
-  binding {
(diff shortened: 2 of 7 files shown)


aiwilliams (approved):
It seems to me this is an exceptional practice. How about specifying the major version of Terraform as well? I use `asdf` to manage my terraform versions, and an entry in `.tool-versions` to make my life easier. 😁

austinkelleher:
@aiwilliams Agreed, but I think we should use the `required_version` constraint instead: https://www.terraform.io/docs/language/settings/index.html#specifying-a-required-terraform-version

aiwilliams:
> @aiwilliams Agreed, but I think we should use the `required_version` constraint instead: https://www.terraform.io/docs/language/settings/index.html#specifying-a-required-terraform-version

Oh, yeah, totally meant that we should add the constraint to the terraform config. I should have said, "I also use `asdf` to manage my installed and selected terraform version."

ndowmon (dismissed):
Interesting. Azure actually allows you to specify the version _on_ the [provider](https://github.com/JupiterOne/graph-azure/blob/master/terraform/main.tf#L1):

```
provider "azurerm" {
  version = "~>2.43.0"
  features {
    key_vault {
      recover_soft_deleted_key_vaults = true
      purge_soft_delete_on_destroy = true
    }
  }
}
```

But this is great to know.

.gitignore
```diff
@@ -1,7 +1,8 @@
 node_modules/
 dist/
 .j1-integration/
-.env
+.env*
 .eslintcache
 tsconfig.tsbuildinfo
 coverage/
+main.remote.tf
```
austinkelleher, line 8: This file will be used by integration developers to target a specific Terraform Cloud workspace. Example:

```terraform
terraform {
  backend "remote" {
    organization = "my-tf-cloud-org"

    workspaces {
      name = "my-tf-org-workspace"
    }
  }
}
```

terraform/cloud_storage_buckets.tf
```diff
@@ -10,60 +10,6 @@ resource "google_kms_crypto_key" "bucket_key" {
   depends_on = [google_kms_key_ring.bucket_key_ring]
 }
 
-resource "google_project_iam_member" "grant_google_storage_service_encrypt_decrypt" {
```
austinkelleher, line 13: Decided that none of these old resources were necessary anymore and were also causing various TF permission issues.

terraform/cloud_storage_buckets.tf
```diff
@@ -10,60 +10,6 @@ resource "google_kms_crypto_key" "bucket_key" {
...
-  bucket = google_storage_bucket.public_storage_bucket.name
-  policy_data = data.google_iam_policy.public_storage_object_viewer.policy_data
-}
-################################################################################
-# End public bucket code
-################################################################################
-
 resource "google_storage_bucket" "sink-log-bucket" {
-  name = "my-unique-sink-logging-bucket"
+  name = "${var.project_id}-sink-logging-bucket"
```
austinkelleher, line 14: Needs to be globally unique

terraform/main.tf
```diff
@@ -1,4 +1,6 @@
 terraform {
+  required_version = "~> 0.14"
```
austinkelleher, line 2: 🥳

terraform/main.tf
```diff
@@ -18,7 +20,7 @@ terraform {
 }
 
 provider "google" {
-  credentials = file(abspath(var.service_account_key_file_path))
+  credentials = var.service_account_key
```
austinkelleher, line 23: Now inside of Terraform Cloud, but could still be specified locally if the user did not have a Terraform Cloud account.

aiwilliams (approved):

PR #81, opened 2021-02-23
Use config path for gitleaks.

austinkelleher:


.gitleaks.toml
```diff
@@ -107,5 +104,5 @@ title = "gitleaks config"
 	files = [
 		'''(.*)?gitleaks.toml$''',
 		'''(.*)?yarn.lock$''',
-		'''(.*?)(jpg|gif|doc|pdf|bin)$)'''
```
austinkelleher, line 110: Gitleaks didn't like this one anymore! Was causing parsing issues. Breaking change?

charlieduong94: oof, is there a way to lock gitleaks versions?

charlieduong94 (approved):

.gitleaks.toml
austinkelleher: LITERALLY ANYTHING IS POSSIBLE IF YOU BELIEVE IN YOURSELF

PR #103, opened 2021-04-12
Remove raw data collection for compute instance entities

austinkelleher:


ndowmon (approved):

src/steps/compute/converters.ts
```diff
@@ -294,7 +294,6 @@ export function createComputeInstanceEntity(data: compute_v1.Schema$Instance) {
   return createGoogleCloudIntegrationEntity(data, {
     entityData: {
       source: {
-        ...data,
         tags: buildComputeInstanceTags(data),
```
ndowmon, line 297: 💯

PR #131, opened 2021-05-05
Add hasIamPermissions property to google_cloud_api_service

austinkelleher:
> Which enabled Google Cloud API services do not use cloud IAM permissions?

```
find google_cloud_api_service with 
  enabled=true and 
  hasIamPermissions=false
return
  google_iam_role.name,
  google_iam_role.displayName,
  google_iam_role.description,
  google_iam_role.permissions,
  google_cloud_api_service.name,
  google_cloud_api_service.displayName,
  google_cloud_api_service.description
```


src/steps/service-usage/index.test.ts
```diff
@@ -1,3 +1,5 @@
+jest.setTimeout(30000);
```
kenanwarren, line 1: Is this jest timeout needed?

kenanwarren (commented):
lgtm other than the small question

src/steps/service-usage/index.test.ts
austinkelleher: Good question. This was necessary when running `test:env`. Some tests are really slow when generating recording files. The actual test shouldn't take nearly this long. We've seen this in a few integrations and I think we need a better strategy. Perhaps we should raise the global timeout to 30000 when running `test:env` by reading an environment variable or something.

kenanwarren (approved):

PR #132, opened 2021-05-05
Add readonly property to google_iam_role

austinkelleher:
This is a bit of a clever method in order to answer the following questions without complex J1QL filtering:

> Which Google Cloud roles are read-only?

```
find google_iam_role with readonly=true
```

> Which Google Cloud roles have write access?

```
find google_iam_role with readonly=false
```


aiwilliams (approved):
Excellent.

kenanwarren (approved):

PR #133, opened 2021-05-06
Build google_cloud_api_service HAS google_iam_role relationship

austinkelleher:
Example queries:

> Which enabled Google Cloud API services are used by roles with read-only permissions?

```
find google_cloud_api_service with enabled=true
THAT HAS google_iam_role with readonly=true
return
  google_iam_role.name,
  google_iam_role.displayName,
  google_iam_role.description,
  google_iam_role.permissions,
  google_cloud_api_service.name,
  google_cloud_api_service.displayName,
  google_cloud_api_service.description
```

> Which enabled Google Cloud API services are used by roles with write permissions?

```
find google_cloud_api_service with enabled=true
THAT HAS google_iam_role with readonly!=true
return
  google_iam_role.name,
  google_iam_role.displayName,
  google_iam_role.description,
  google_iam_role.permissions,
  google_cloud_api_service.name,
  google_cloud_api_service.displayName,
  google_cloud_api_service.description
```


src/steps/service-usage/converters.ts
```diff
@@ -21,7 +36,10 @@ export function createApiServiceEntity({
       assign: {
         _class: API_SERVICE_ENTITY_CLASS,
         _type: API_SERVICE_ENTITY_TYPE,
-        _key: data.name as string,
+        _key: getServiceApiEntityKey({
```
austinkelleher, line 39: I had to change the `_key` value to make it possible to easily find it later.

src/steps/iam/index.ts
```diff
@@ -18,26 +24,116 @@ import {
...
+} from '../../utils/iam';
+import {
+  API_SERVICE_ENTITY_TYPE,
+  STEP_API_SERVICES,
+} from '../service-usage/constants';
+import { getServiceApiEntityKey } from '../service-usage/converters';
 
 export * from './constants';
 
+async function buildApiServiceEntityHasIamRoleRelationships({
```
mknoedel, line 43: Is `build` standard here? seems more like a `create` as we are adding stuff to the jobstate.

mknoedel (dismissed):
Looks great!

src/steps/iam/index.ts
```diff
@@ -18,26 +24,116 @@ import {
...
+            serviceApiName,
+          },
+          'Could not find service API entity when attempting to build relationship to role',
+        );
+        serviceApiEntityKeySeenMap.set(serviceApiEntityKey, false);
+        continue;
+      }
+
+      serviceApiEntityKeySeenMap.set(serviceApiEntityKey, true);
+    } else if (serviceApiEntityKeySeenMapValue === false) {
```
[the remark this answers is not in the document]
austinkelleher: Ya this is a bit confusing. This is sort of a special case. `undefined` means we haven't looked up the data at all yet. `false` means we actually tried to do a lookup and it didn't exist. It never will exist, so we are caching the fact that it will _not_ exist to prevent duplicate lookups later.

mknoedel (approved):
Great changes thanks!

PR #183, opened 2021-05-25
Update `google_bigquery_table` `_class` to `DataCollection`

austinkelleher:
FYI @eXtremeX. We decided to introduce a new class called `DataCollection`. We had considered `DataTable`, but that makes assertions about the structure of the data (in table form).

> What tables do my databases use?

```
FIND Database
THAT HAS DataCollection
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,23 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Added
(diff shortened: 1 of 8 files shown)


aiwilliams (approved):

eXtremeX:
Understood, thanks for letting me know!

PR #187, opened 2021-05-27
Fixes #186 - Accept `storage-api` and `storage-component` services to enable buckets step

### LINKED ISSUE #186: Accept `storage-api` and `storage-component` services to enable buckets step
These APIs can be enabled to fetch buckets

austinkelleher:
#186

### DIFF
--- a/src/getStepStartStates.ts
+++ b/src/getStepStartStates.ts
@@ -138,11 +138,24 @@ export default async function getStepStartStates(
   }
 
   const createStepStartState = (
-    serviceName: ServiceUsageName,
+    primaryServiceName: ServiceUsageName,
+    ...additionalServiceNames: ServiceUsageName[]
   ): StepStartState => {
-    return {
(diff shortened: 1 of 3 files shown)


src/getStepStartStates.ts
```diff
@@ -138,11 +138,24 @@ export default async function getStepStartStates(
...
+      primaryServiceName,
+      ...additionalServiceNames,
+    ];
+
+    let disabled = true;
+
+    for (const serviceName of allServicesToEnableStep) {
+      if (enabledServiceNames.includes(serviceName)) {
+        disabled = false;
+        continue;
```
mknoedel, line 154: this should probably be break

mknoedel (approved):

src/getStepStartStates.ts
austinkelleher: lolz oops

PR #189, opened 2021-05-29
Additional config properties for org auto-config

austinkelleher:


src/index.ts
```diff
@@ -71,3 +73,9 @@ export const invocationConfig: IntegrationInvocationConfig<IntegrationConfig> =
       };
     },
   };
+
+export {
```
austinkelleher, line 77: Using these in managed GCP integration

src/types.ts
```diff
@@ -9,6 +9,19 @@ export type IntegrationStepContext =
 
 export interface SerializedIntegrationConfig extends IntegrationInstanceConfig {
```
austinkelleher, line 10: Added these properties for org auto-config

package.json
```diff
@@ -48,6 +48,7 @@
   "dependencies": {
     "@lifeomic/attempt": "^3.0.0",
     "gaxios": "^4.2.1",
-    "googleapis": "^73.0.0"
+    "google-auth-library": "^7.1.0",
```
austinkelleher, line 51: This package was being used and should have been a direct dep. It was technically being installed as a transitive dep

src/google-cloud/client.ts
```diff
@@ -29,12 +23,6 @@ export type PageableGaxiosResponse<T> = GaxiosResponse<
   }
 >;
 
-export type GoogleClientAuth =
```
austinkelleher, line 32: The Google auth client doesn't export the TS type for this. I did some casting below to simplify things a bit even though it's ugly and sucks.

aiwilliams (approved):

PR #193, opened 2021-05-30
Fixes #192 - Intersection of target project and main project service enablement

### LINKED ISSUE #192: Improve `getStepStartStates` to consider whether target project ID
We need to consider whether both the target project ID and the main project ID have API services enabled. This should be the intersection of the two.

austinkelleher:


src/steps/enablement.test.ts
```diff
@@ -0,0 +1,150 @@
...
+      projectId: 'my-target-project-id',
+    });
+
+    expect(
+      enablement.getUniqueIntegrationConfigProjectsForStepEnablement(config),
+    ).toEqual(['j1-gc-integration-dev-v2', 'my-target-project-id']);
+  });
+});
+
+// describe('#serviceResourceNameToServiceName', () => {
```
austinkelleher, line 25: Whoops I moved this and forgot to add back the test. Fixing...

aiwilliams (approved):

PR #234, opened 2021-06-09
Improve job log event and disable unused Terraform

austinkelleher:


terraform/iam.tf
```diff
@@ -8,17 +8,17 @@ resource "google_project_iam_custom_role" "ga_custom_role" {
...
-  title       = "Security Reviewer Extender Role"
-  description = "A role that adds necessary permissions for the integration that aren't granted via Security Reviewer role"
-  permissions = [
-    "compute.projects.get",
-    "resourcemanager.projects.get",
-    "binaryauthorization.policy.get",
-    "appengine.applications.get",
-  ]
-}
+# resource "google_project_iam_custom_role" "extended_role" {
```
austinkelleher, line 11: We are now configuring this at the org-level. No longer needed, but keeping the code around for reference.

src/getStepStartStates.ts
```diff
@@ -144,7 +144,9 @@ export default async function getStepStartStates(
 
   logger.publishEvent({
     name: 'integration_config',
-    description: `Starting Google Cloud integration with service account (email=${config.serviceAccountKeyConfig.client_email})`,
+    description: `Starting Google Cloud integration with service account (email=${
+      config.serviceAccountKeyConfig.client_email
+    }, configureOrganizationProjects=${!!config.configureOrganizationProjects})`,
```
austinkelleher, line 149: This is the main change

ndowmon (approved):

PR #240, opened 2021-06-10
Fixes #239 - google_iam_role should assign the actual target project projectId instead of the org project

### LINKED ISSUE #239: `google_iam_role` should assign the actual target project `projectId` instead of the org project

austinkelleher:


kenanwarren (approved):

kenanwarren (approved):

src/steps/cloud-asset/index.ts
```diff
@@ -34,8 +36,20 @@ export async function fetchIamBindings(
...
 
+        let projectId: string | undefined;
+
+        if (projectName) {
+          projectId = await getProjectIdFromName(jobState, projectName);
+
+          if (!projectId) {
+            // This should never happen because this step depends on another
+            // step that collects all of the project data in an organization.
+            logger.warn({ projectName }, 'Missing project ID in local cache');
```
aiwilliams, line 47: If this should never happen, would it be better to throw an error that prevents further processing, or perhaps log as `error` so that we can more readily notice in logs (`level=50`)?

Issue #248, opened 2021-06-22
Always execute steps in org child projects even if child project has service disabled


austinkelleher:
We need to do this because some steps have `dependsOn` of other steps. We may miss data in child projects if we don't do this.

See: https://github.com/JupiterOne/graph-google-cloud/blob/67cd5b83c63d81d759ca9b302463a856b90006b7/src/steps/enablement.ts#L33

PR #249, opened 2021-06-22
Always execute steps in org child projects even if child project has service disabled

austinkelleher:
See #248

### DIFF
--- a/src/steps/enablement.ts
+++ b/src/steps/enablement.ts
@@ -4,32 +4,6 @@ import { IntegrationConfig } from '../types';
 import { collectEnabledServicesForProject } from './service-usage/client';
 
 /**
- * An integration can provide a service account key file for a service account
- * that exists in one project, but has permissions to ingest data from a
- * _different_ project. This function determines which unique projects should
- * be considered when calculating step enablement.
- *
- * The first index in the array is guaranteed to be the "main" project
- */
-export function getUniqueIntegrationConfigProjectsForStepEnablement(
-  config: IntegrationConfig,
-): string[] {
-  const uniqueProjectIds: string[] = [
-    // Main project ID
-    config.serviceAccountKeyConfig.project_id,
-  ];
-
-  if (
-    config.projectId &&
-    config.projectId !== config.serviceAccountKeyConfig.project_id
-  ) {
-    uniqueProjectIds.push(config.projectId);
-  }
-
-  return uniqueProjectIds;
-}
-
-/**
  * Determines the overall services that are enabled. The result information is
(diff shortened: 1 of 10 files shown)


src/steps/functions/index.ts
```diff
@@ -26,44 +27,67 @@ export async function fetchCloudFunctions(
...
+      );
+    },
+  );
 }
 
 export const functionsSteps: IntegrationStep<IntegrationConfig>[] = [
   {
     id: STEP_CLOUD_FUNCTIONS,
     name: 'Cloud Functions',
-    dependsOn: [STEP_IAM_SERVICE_ACCOUNTS],
```
austinkelleher, line 59: These are bonus changes that are somewhat related. This issue manifested itself because the cloud function step was not executing. It turns out that the step enablement logic was disabling the `fetch-iam-service-accounts` step, was not being executed.

src/steps/functions/converters.ts
```diff
@@ -36,6 +36,7 @@ export function createCloudFunctionEntity(
         maxInstances: cloudFunction.maxInstances,
         status: cloudFunction.status,
         active: cloudFunction.status === 'ACTIVE',
+        serviceAccountEmail: cloudFunction.serviceAccountEmail,
```
austinkelleher, line 39: Another small bonus change

src/steps/functions/index.ts
mknoedel: So do you need the other changes in this PR if this was causing the issue?

src/steps/enablement.ts
```diff
@@ -39,48 +13,15 @@ export function getUniqueIntegrationConfigProjectsForStepEnablement(
...
- * can speed up the integration execution by simply disabling steps if the
- * "target" project has the APIs disabled, but the "main" project has those APIs
- * enabled.
+ * project does not have that API enabled, the integration can hit the APIs.
  */
 export async function getEnabledServiceNames(
   config: IntegrationConfig,
 ): Promise<string[]> {
-  const [mainProjectId, targetProjectId] =
-    getUniqueIntegrationConfigProjectsForStepEnablement(config);
```
mknoedel, line 53: Was this change mostly for safety and because there isn't really a problem with hitting apis with nothing provisioned in them?

mknoedel (approved):

src/steps/functions/index.ts
austinkelleher: Yes because there are other steps in the codebase that should be split out like this. I just addressed one of them.

mknoedel (approved):

PR #251, opened 2021-06-23
Update org setup instructions and remove old CLI instructions

austinkelleher:


docs/jupiterone.md
```diff
@@ -216,104 +216,6 @@ organization, the auto-configuration process may take a few minutes to complete.
 When the process has been completed, you will see your new integration instances
 on the JupiterOne Google Cloud integration list page.
 
-#### JupiterOne + Google Cloud Organization CLI
```
austinkelleher, line 219: We shouldn't tell clients to use this anymore. Instead, clients should use the J1 UI approach.

ndowmon: Do you want to actually _remove_ the script from this project as well?

ndowmon (approved):

docs/jupiterone.md
austinkelleher: Probably, yes.

PR #261, opened 2021-07-14
google_iam_binding_uses_role

mknoedel:
Direct:
image

Mapped: 
image


src/steps/cloud-asset/index.ts
```diff
@@ -48,10 +60,48 @@ export async function fetchIamBindings(
...
+                  IAM_ROLE_ENTITY_TYPE,
+                ),
+                properties: {
+                  projectId,
+                },
+                _mapping: {
+                  relationshipDirection: RelationshipDirection.FORWARD,
+                  sourceEntityKey: bindingEntity._key,
+                  targetFilterKeys: [['_type', '_key']],
+                  skipTargetCreation: false,
```
mknoedel, line 94: Do we want to skip target creation here? from the looks of it, the vast majority of iam_roles are not ingested by this integration. I think they are google managed or something like that.

src/steps/cloud-asset/index.ts
```diff
@@ -48,10 +60,48 @@ export async function fetchIamBindings(
           }
         }
 
-        await jobState.addEntity(
+        const bindingEntity = await jobState.addEntity(
           createIamBindingEntity({ _key, projectId, binding, resource }),
         );
 
+        if (binding.role) {
```
mknoedel, line 67: Do we need to account for project-level role names differently (ex: `projects/j1-gc-integration-dev-v3/roles/167984947943customrole`)? Not that I could tell from our test environment, but it was something that stood out as other bindings looked to be `groups/admin` and stuff like that.

src/steps/cloud-asset/index.ts
```diff
@@ -48,10 +60,48 @@ export async function fetchIamBindings(
...
+            await jobState.addRelationship(
+              createMappedRelationship({
+                _class: RelationshipClass.USES,
+                _type: generateRelationshipType(
+                  RelationshipClass.USES,
+                  bindingEntities.BINDINGS._type,
+                  IAM_ROLE_ENTITY_TYPE,
+                ),
+                properties: {
+                  projectId,
```
austinkelleher, line 88: We should probably remove this. I think that this role could be a custom role that exists as part of a child project that will get picked up later by another integration instance.

src/steps/cloud-asset/index.ts
austinkelleher: Similar to above, I think we should create this target entity. This is likely a "custom" role owned by another child project integration instance.

src/steps/cloud-asset/index.ts
```diff
@@ -48,10 +60,48 @@ export async function fetchIamBindings(
...
-        await jobState.addEntity(
+        const bindingEntity = await jobState.addEntity(
           createIamBindingEntity({ _key, projectId, binding, resource }),
         );
 
+        if (binding.role) {
+          const roleEntity = await jobState.findEntity(binding.role);
+
+          if (roleEntity) {
+            await jobState.addRelationship(
```
austinkelleher, line 71: Please make sure that we capture direct relationships to "managed" role entities.

austinkelleher (changes requested):

src/steps/cloud-asset/index.ts
mknoedel: Done by pulling this out into it's own step and then making it dependent on the role-fetching and creating steps.

austinkelleher (approved):
Amazing stuff! Super excited about these changes!

PR #265, opened 2021-07-16
INT-1155 binding assigned principal

mknoedel:
Screen Shot 2021-07-15 at 3 51 15 PM
Screen Shot 2021-07-16 at 1 45 01 PM
Screen Shot 2021-07-16 at 1 46 43 PM


docs/jupiterone.md
```diff
@@ -324,7 +324,7 @@ The following entities are created:
 | DNS Managed Zone                                         | `google_dns_managed_zone`                                         | `DomainZone`                       |
 | Folder                                                   | `google_cloud_folder`                                             | `Group`                            |
 | IAM Binding                                              | `google_iam_binding`                                              | `AccessPolicy`                     |
-| IAM Managed Role                                         | `google_iam_role`                                                 | `AccessRole`                       |
+| IAM Custom Role                                          | `google_iam_role`                                                 | `AccessRole`                       |
```
mknoedel, line 327: This change was due to incorrectly stating that the `IAM Managed Role` step was creating `google_iam_role` entities, which it was not, the `IAM Custom Role` and `IAM Bindings` do however but the documentation tool only allows for one step.

docs/jupiterone.md
```diff
@@ -413,6 +413,10 @@ The following relationships are created/mapped:
 | `google_container_node_pool`                                     | **HAS**               | `google_compute_instance_group`                                   |
 | `google_cloud_folder`                                            | **HAS**               | `google_cloud_project`                                            |
 | `google_group`                                                   | **ASSIGNED**          | `google_iam_role`                                                 |
+| `google_iam_binding`                                             | **ASSIGNED**          | `google_group`                                                    |
+| `google_iam_binding`                                             | **ASSIGNED**          | `google_iam_service_account`                                      |
+| `google_iam_binding`                                             | **ASSIGNED**          | `google_user`                                                     |
+| `google_iam_binding`                                             | **USES**              | `google_iam_role`                                                 |
```
mknoedel, line 419: Forgot to add this in the previous PR

src/getStepStartStates.ts
```diff
@@ -229,9 +229,9 @@ export default async function getStepStartStates(
     [STEP_IAM_CUSTOM_ROLES]: createStepStartState(ServiceUsageName.IAM),
     [STEP_IAM_MANAGED_ROLES]: createStepStartState(ServiceUsageName.IAM),
     [STEP_IAM_SERVICE_ACCOUNTS]: createStepStartState(ServiceUsageName.IAM),
-    [STEP_RESOURCE_MANAGER_IAM_POLICY]: createStepStartState(
-      ServiceUsageName.RESOURCE_MANAGER,
-    ),
+    [STEP_RESOURCE_MANAGER_IAM_POLICY]: config.configureOrganizationProjects
+      ? { disabled: true }
+      : createStepStartState(ServiceUsageName.RESOURCE_MANAGER),
```
mknoedel, line 234: I believe we don't want to run this step when we are ingesting organization data.

The problem is that both `fetch-resource-manager-iam-policy` and `fetch-iam-bindings` ingest some roles, and roles => principal relationships, but `fetch-iam-bindings` ingests far more data (along with binding entities and their relationships) but requires Organization level permissions. Because of this, I think we don't want to run `fetch-resource-manager-iam-policy` in any google organization that has access to `fetch-iam-bindings`. The way I think this can be done is by checking the `configureOrganizationProjects` config variable.

src/getStepStartStates.ts
```diff
@@ -212,9 +215,9 @@ export default async function getStepStartStates(
     // This API will be enabled otherwise fetching services names above would fail
     [STEP_RESOURCE_MANAGER_PROJECT]: { disabled: false },
     [STEP_API_SERVICES]: { disabled: false },
-    [CLOUD_ASSET_STEPS.BINDINGS]: createOrgStepStartState(
-      ServiceUsageName.CLOUD_ASSET,
-    ),
+    [CLOUD_ASSET_STEPS.BINDINGS]: !isMasterOrgInstance
+      ? { disabled: true }
+      : createStepStartState(ServiceUsageName.CLOUD_ASSET),
```
mknoedel, line 220: (read the lower comment in this file first for context) Similar to my comment below, I think we only want to enable this step for one integration instance instead of all the project instances that get automatically created. 

This step also requires organization level permissions so we shouldn't try to run it on instances that aren't using `configureOrganizationProjects` I believe.

src/getStepStartStates.ts
```diff
@@ -212,9 +215,9 @@ export default async function getStepStartStates(
     // This API will be enabled otherwise fetching services names above would fail
     [STEP_RESOURCE_MANAGER_PROJECT]: { disabled: false },
     [STEP_API_SERVICES]: { disabled: false },
-    [CLOUD_ASSET_STEPS.BINDINGS]: createOrgStepStartState(
-      ServiceUsageName.CLOUD_ASSET,
-    ),
+    [STEP_IAM_BINDINGS]: !isMasterOrgInstance
+      ? { disabled: true }
+      : createStepStartState(ServiceUsageName.CLOUD_ASSET),
```
mknoedel, line 220: (read the lower comment in this file first for context) Similar to my comment below, I think we only want to enable this step for one integration instance instead of all the project instances that get automatically created.

This step also requires organization level permissions so we shouldn't try to run it on instances that aren't using configureOrganizationProjects I believe.

src/steps/cloud-asset/constants.ts
```diff
@@ -16,40 +20,54 @@ export const bindingEntities = {
   },
 };
 
-const IAM_TARGET_TYPES = [
-  'google_domain',
-  'google_iam_service_account',
-  'google_group',
-  'google_user',
+const IAM_PRINCIPAL_TYPES = [
+  // 'google_domain', // These are getting ignored for now
```
mknoedel, line 24: I wasn't getting any of these and we weren't handling these cases in the `STEP_RESOURCE_MANAGER_IAM_POLICY` step, but [the google documentation](https://cloud.google.com/asset-inventory/docs/reference/rpc/google.iam.v1#google.iam.v1.Binding) says there is another level that represents "all the users of that domain. For example, google.com or example.com." 

I'm not sure how common these are, but skipping for this PR at least.

src/steps/cloud-asset/constants.ts
```diff
@@ -16,40 +20,54 @@ export const bindingEntities = {
...
+        RelationshipClass.ASSIGNED,
+        principalType,
+        IAM_ROLE_ENTITY_TYPE,
+      ),
+      sourceType: principalType,
+      _class: RelationshipClass.ASSIGNED,
+      targetType: IAM_ROLE_ENTITY_TYPE,
+    };
   },
+);
```
mknoedel, line 58: In order to maintain backwards compatibility with `STEP_RESOURCE_MANAGER_IAM_POLICY` I'm creating role -> principal relationships here as well.

src/steps/cloud-asset/index.test.ts
```diff
@@ -9,33 +9,100 @@ import {
...
       await fetchIamBindings(context);
 
       expect({
         numCollectedEntities: context.jobState.collectedEntities.length,
         numCollectedRelationships:
           context.jobState.collectedRelationships.length,
-        collectedEntities: context.jobState.collectedEntities,
-        collectedRelationships: context.jobState.collectedRelationships,
+        collectedEntities: context.jobState.collectedEntities.length,
+        collectedRelationships: context.jobState.collectedRelationships.length,
```
mknoedel, line 66: I don't think there is much value in storing hundreds of graph objects when the purpose of this test is higher level than that.

src/steps/cloud-asset/index.ts
```diff
@@ -64,6 +79,9 @@ export async function fetchIamBindings(
           createIamBindingEntity({ _key, projectId, binding, resource }),
         );
 
+        bindingGraphKeySet.add(_key);
+        iamBindingsCount++;
+
```
mknoedel, line 84: Safer to have this before any relationship creation I think

src/steps/cloud-asset/index.ts
```diff
@@ -102,8 +120,62 @@ export async function fetchIamBindings(
           }
         }
 
-        bindingGraphKeySet.add(_key);
-        iamBindingsCount++;
+        if (binding.members?.length) {
```
mknoedel, line 123: In a future PR, I think I'm going to break out each of these relationship branches to their own step, this one is getting quite large.

src/steps/cloud-asset/index.ts
```diff
@@ -142,14 +210,35 @@ export async function fetchIamBindings(
...
     name: 'IAM Bindings',
-    entities: [bindingEntities.BINDINGS],
-    relationships: [],
+    entities: [
+      bindingEntities.BINDINGS,
+      {
+        resourceName: 'IAM Role',
+        _type: IAM_ROLE_ENTITY_TYPE,
+        _class: IAM_ROLE_ENTITY_CLASS,
+      },
```
mknoedel, line 221: Comes from `findOrCreateIamRoleEntity` and is how this previously functioned. I believe we need to create these roles because we need to build mapped relationships off of this entity which is currently not supported by the persister. This may be causing duplicate role entities today though I'm not sure.

src/steps/iam/index.ts
```diff
@@ -215,13 +215,7 @@ export const iamSteps: IntegrationStep<IntegrationConfig>[] = [
...
     id: STEP_IAM_MANAGED_ROLES,
     name: 'Identity and Access Management (IAM) Managed Roles',
-    entities: [
-      {
-        resourceName: 'IAM Managed Role',
-        _type: IAM_ROLE_ENTITY_TYPE,
-        _class: IAM_ROLE_ENTITY_CLASS,
-      },
-    ],
+    entities: [],
```
mknoedel, line 218: This step only added data to the jobState, it didn't create entities.

src/steps/resource-manager/converters.ts
```diff
@@ -35,35 +33,19 @@ function getConditionRelationshipProperties(
   };
 }
 
-export function createIamServiceAccountAssignedIamRoleRelationship(params: {
```
mknoedel, line 38: I deleted this method and it's associated tests because it was more confusing than it was worth I think.

src/steps/resource-manager/converters.ts
```diff
@@ -35,35 +33,19 @@ function getConditionRelationshipProperties(
...
-    to: params.iamRoleEntity,
-    properties: {
-      projectId: params.projectId,
-      ...(params.condition &&
-        getConditionRelationshipProperties(params.condition)),
-    },
-  });
-}
-
 export function createGoogleWorkspaceEntityTypeAssignedIamRoleMappedRelationship({
```
mknoedel, line 36: Updated this method to work with both roles and bindings, the relationship direction stuff made it kind of dirty which I'm not happy about, but I couldn't think of a better way a the time.

src/steps/resource-manager/index.ts
```diff
@@ -150,67 +148,70 @@ async function findOrCreateIamRoleEntity({
...
-  data,
+  condition,
+  relationshipDirection,
 }: {
-  jobState: JobState;
-  projectId: string;
-  data: PolicyMemberBinding;
-}): Promise<Relationship | undefined> {
-  const roleName = data.binding.role as string;
-  const iamUserEntityWithParsedMember =
```
mknoedel, line 163: Took this function call out of this method to make it more pure and easier to reuse the conditional logic below.

src/steps/resource-manager/index.ts
```diff
@@ -150,67 +148,70 @@ async function findOrCreateIamRoleEntity({
...
+  condition?: cloudresourcemanager_v3.Schema$Expr;
+  relationshipDirection?: RelationshipDirection;
+}): Relationship | undefined {
+  relationshipDirection = relationshipDirection
+    ? relationshipDirection
+    : RelationshipDirection.REVERSE;
   if (iamUserEntityWithParsedMember.userEntity) {
     // Create a direct relationship. This is a user entity that _only_ exists
     // in the Google Cloud integration (e.g. a GCP service account)
-    const iamRoleEntity = await findOrCreateIamRoleEntity({
```
mknoedel, line 176: Same with this method.

src/steps/resource-manager/index.ts
```diff
@@ -394,23 +395,46 @@ export async function fetchResourceManagerIamPolicy(
...
 
   await client.iteratePolicyMemberBindings(async (data) => {
-    const relationship = await buildIamUserRoleRelationship({
-      jobState,
-      projectId: client.projectId,
-      data,
-    });
-
-    if (!relationship || relationships.has(relationship._key)) {
+    if (!data.binding.role) {
```
mknoedel, line 404: This was needed once we upgraded the typings.

mknoedel, line 404: You can't build relationships to a role that doesn't exist.

src/steps/cloud-asset/index.test.ts
```diff
@@ -9,33 +9,100 @@ import {
...
+        google_iam_binding_assigned_service_account,
+        google_user_assigned_iam_role,
+        google_group_assigned_iam_role,
+        google_iam_service_account_assigned_role,
+      } = separateGraphObjectsByType(
+        context.jobState.collectedRelationships,
+        context.jobState.encounteredTypes,
+      );
+
+      // Each of these relationships could either be mapped or direct depending on if it has been previously ingested in this run so not checking if direct or mapped here
```
austinkelleher, line 84: It would be great if we could add an assertion for both mapped and non-mapped, and that our recordings would include both cases to prevent later regression. Thanks for raising!

austinkelleher (changes requested):
Nice job!

src/steps/cloud-asset/index.test.ts
mknoedel: I made the scaffolding but didn't have examples right off the bat for these. I'll add these examples in a later PR when I get more familiar with the GCP platform.

austinkelleher (approved):
Thank you for all of your work on this!

PR #272, opened 2021-07-22
binding to google_domain relationships

mknoedel:
google domain entity and relationships
Screen Shot 2021-07-22 at 10 52 12 AM

data changes for google group and user mapped entities
Screen Shot 2021-07-22 at 10 53 01 AM
Screen Shot 2021-07-22 at 10 51 51 AM


src/steps/app-engine/__snapshots__/index.test.ts.snap
```diff
@@ -9676,16 +9676,16 @@ Object {
         "skipTargetCreation": false,
         "sourceEntityKey": "roles/owner",
         "targetEntity": Object {
+          "_class": "User",
           "_type": "google_user",
           "deleted": false,
-          "displayName": "austin.kelleher@jupiterone.dev",
           "email": "austin.kelleher@jupiterone.dev",
-          "name": "austin.kelleher@jupiterone.dev",
```
mknoedel, line 9683: The `name` and `displayName` properties are set to a different property in `graph-google` and I didn't want to overwrite correct name information: ex: "Michael Knoedel, with incorrect name information: ex: "michael.knoedel@jupiterone.com"

src/steps/cloud-asset/index.ts
```diff
@@ -231,6 +231,7 @@ export async function createPrincipalRelationships(
                     projectId: bindingEntity.projectId,
                     iamUserEntityWithParsedMember,
                     condition,
+                    relationshipDirection: RelationshipDirection.REVERSE,
```
mknoedel, line 234: Made this key required

src/steps/resource-manager/converters.ts
```diff
@@ -34,67 +33,46 @@ export function getConditionRelationshipProperties(
...
-  iamUserEntityWithParsedMember,
+  targetEntity,
   relationshipDirection,
   projectId,
   condition,
 }: {
-  targetEntityType: 'google_group' | 'google_user';
   iamEntity: Entity;
-  iamUserEntityWithParsedMember: IamUserEntityWithParsedMember;
+  targetEntity: Partial<PrimitiveEntity>;
```
mknoedel, line 43: Pulled out targetEntity creation logic out of this method in order to simplify this method a bit.

src/steps/resource-manager/converters.ts
```diff
@@ -34,67 +33,46 @@ export function getConditionRelationshipProperties(
...
-    });
-  }
-
   const iamEntityToTargetRelationship =
     relationshipDirection === RelationshipDirection.FORWARD;
 
+  // Not always able to determine a _key for google_users depending on how the binding is set up
+  const targetFilterKeys = targetEntity._key
+    ? [['_key']]
+    : [['_type', '_class', 'email']];
```
mknoedel, line 54: Optimization to only search on what you know the entity has access to.

src/steps/resource-manager/converters.ts
```diff
@@ -34,67 +33,46 @@ export function getConditionRelationshipProperties(
...
+  const targetKey = targetEntity._key ?? (targetEntity.email as string);
```
mknoedel, line 55: The closest thing we will have to a key is an email which is what we had used before, though part of me thinks it would be better to just let the mapper make the relationship key instead of passing in the email.

I believe if you don't give a _key property to a mapped relationship, the key of the target entity becomes `mapped_entity` or something like that.

src/steps/resource-manager/converters.ts
```diff
@@ -34,67 +33,46 @@ export function getConditionRelationshipProperties(
...
-         * google_groups have their own "name" identifier which we do not have access via the bindings.
-         * Therefore, for a google_user, we set their `name` and `displayName properties to the email per
-         * https://github.com/JupiterOne/graph-google/blob/master/src/steps/groups/converters.ts#L209,
-         * but not for google_groups per
-         * https://github.com/JupiterOne/graph-google/blob/master/src/steps/groups/converters.ts#L39
-         */
-        name: targetEntityType === 'google_user' ? email : undefined,
-        displayName: targetEntityType === 'google_user' ? email : undefined,
-        deleted: iamUserEntityWithParsedMember.parsedMember.deleted,
-      },
```
mknoedel, line 86: This all gets handled on a per-entity type basis now to prevent if statements everywhere.

src/steps/resource-manager/createIamEntities.ts
```diff
@@ -0,0 +1,114 @@
...
+  GOOGLE_USER_ENTITY_TYPE,
+  IAM_PRINCIPAL_TYPE,
+} from '../iam';
+import {
+  IntegrationError,
+  PrimitiveEntity,
+} from '@jupiterone/integration-sdk-core';
+import { IamUserEntityWithParsedMember } from '.';
+
+// https://github.com/JupiterOne/graph-google/blob/ad4d88d0151cd7dc4ad93dc30d1aa40e6c97778e/src/steps/domains/converters.ts#L6
```
mknoedel, line 16: All of these methods were taken from the graph-google integration to ensure I'm not corrupting entity data if the mapped relationships hit existing entities.

src/steps/app-engine/__snapshots__/index.test.ts.snap
```diff
@@ -9676,16 +9676,16 @@ Object {
...
-          "username": "austin.kelleher@jupiterone.dev",
+          "username": "austin.kelleher",
         },
         "targetFilterKeys": Array [
           Array [
             "_type",
+            "_class",
```
aiwilliams, line 9688: I think it would be good to drop this because `_type: 'google_user'` is very specific, and not including `_class` makes it so that we can change/add/remove values in the future without consequence (in terms of target resolution).

src/steps/resource-manager/converters.ts
```diff
@@ -34,67 +33,46 @@ export function getConditionRelationshipProperties(
...
-      code: 'UNPROCESSABLE_GOOGLE_WORKSPACE_MAPPED_RELATIONSHIP',
-    });
-  }
-
   const iamEntityToTargetRelationship =
     relationshipDirection === RelationshipDirection.FORWARD;
 
+  // Not always able to determine a _key for google_users depending on how the binding is set up
+  const targetFilterKeys = targetEntity._key
+    ? [['_key']]
```
aiwilliams, line 53: I think it'd be good to make this `[_type, _key]` so that at some point when we have the optimization of finding a specific, single entity using an index on those two properties, this will benefit.

src/steps/resource-manager/converters.ts
aiwilliams: Can drop `_class` because `_type` is specific enough.

src/steps/resource-manager/converters.ts
aiwilliams: I wonder what the behavior is when we specify a target entity _key and other integrations attempt to do the say, but disagree on what the value should be?

src/steps/resource-manager/createIamEntities.ts
aiwilliams: 🔥

aiwilliams (changes requested):

src/steps/resource-manager/converters.ts
mknoedel: After a long search, it looks like _key and _type values will never be overwritten by the mapper - https://github.com/JupiterOne/jupiter-mapper/blob/main/src/engine/BuildPlan/PropertyMerger.ts#L119

aiwilliams (approved):

PR #279, opened 2021-07-30
Create IAM bindings to cloud projects

mknoedel:
The issue here is that we use the project name as the `google_cloud_project`s key which is not always available in the Google resource identifier. The identifier could represent either the name or the id of the project and it is up to us to figure out which one. Luckily, we ingest every cloud project earlier on in the integration run, so we can just use data from that step as a reference when making the binding mapped relationships to cloud projects.


src/steps/cloud-asset/index.test.ts
```diff
@@ -273,7 +275,11 @@ describe('#fetchIamBindings', () => {
         google_iam_service_account_assigned_role,
         google_iam_binding_allows_cloud_organization,
         google_iam_binding_allows_cloud_folder,
-        google_iam_binding_allows_ANY_RESOURCE,
```
mknoedel, line 276: renamed this key to be specific resources instead of have a type of "ANY_RESOURCE". "ANY_RESOURCE" is still used for documentation though.

src/steps/cloud-asset/index.test.ts
```diff
@@ -298,8 +309,10 @@ describe('#fetchIamBindings', () => {
         'google_group_assigned_iam_role',
       );
       expect(
-        google_iam_binding_allows_ANY_RESOURCE,
-      ).toHaveOnlyMappedRelationships('google_iam_binding_allows_ANY_RESOURCE');
+        google_iam_binding_allows_cloud_project,
+      ).toHaveOnlyMappedRelationships(
+        'google_iam_binding_allows_cloud_project',
```
mknoedel, line 314: projects are always mapped when doing organization ingestion.

src/steps/cloud-asset/__recordings__/createMappedBindingAnyResourceRelationships_3853245014/recording.har
```diff
@@ -2672,6 +2672,249 @@
           "ssl": -1,
           "wait": 404
         }
+      },
+      {
```
mknoedel, line 2676: Don't know why this was added, but it does return data, so I'm just going to leave it in 🙃

src/steps/cloud-asset/index.ts
```diff
@@ -258,7 +258,10 @@ export async function createBindingToAnyResourceRelationships(
       const { type, key } =
         makeLogsForTypeAndKeyResponse(
           logger,
-          getTypeAndKeyFromResourceIdentifier(bindingEntity.resource),
+          await getTypeAndKeyFromResourceIdentifier(
+            bindingEntity.resource,
+            context,
```
mknoedel, line 263: Needed to add context so the keyGen methods could have access to the jobState.

src/utils/iamBindings/findResourceKindFromCloudResourceIdentifier.test.ts
```diff
@@ -1,20 +1,21 @@
...
   [`//cloudfunctions.googleapis.com/projects/${TEST_PROJECT_ID}/locations/${TEST_LOCATION}/functions/${GENERIC_ID}`]:
     'cloudfunctions.googleapis.com/CloudFunction',
   [`//cloudresourcemanager.googleapis.com/organizations/${GENERIC_ID}`]:
     'cloudresourcemanager.googleapis.com/Organization',
   [`//cloudresourcemanager.googleapis.com/folders/${GENERIC_ID}`]:
     'cloudresourcemanager.googleapis.com/Folder',
-  [`//cloudresourcemanager.googleapis.com/projects/${GENERIC_ID}`]:
+  [`//cloudresourcemanager.googleapis.com/projects/${TEST_PROJECT_NAME}`]:
     'cloudresourcemanager.googleapis.com/Project',
   [`//cloudresourcemanager.googleapis.com/projects/${TEST_PROJECT_ID}`]:
```
mknoedel, line 20: Because a project identifier could be either a projectName or a projectId, I needed to make sure that I had a test for both cases.

src/utils/iamBindings/typeToKeyGeneratorMap.ts
```diff
@@ -78,18 +78,27 @@ import {
...
 } = {
   [CLOUD_FUNCTION_ENTITY_TYPE]: fullPathKeyMap,
   [ORGANIZATION_ENTITY_TYPE]: fullPathKeyMap,
   [FOLDER_ENTITY_TYPE]: fullPathKeyMap,
-  [PROJECT_ENTITY_TYPE]: impossible, // // can't map directly to allUniqueIdentifiers right now as we need a way to differentiate between a PROJECT_NAME and a PROJECT_ID
+  [PROJECT_ENTITY_TYPE]: async (id: string, context?: StepExecutionContext) =>
+    (await getProjectNameFromId(
+      context!.jobState,
+      finalIdentifierKeyMap(id),
+    )) ?? finalIdentifierKeyMap(id),
```
mknoedel, line 101: This logic searches the jobState in the `projectId` map to get it's related name. If the identifier is in fact a name, it will not be found in the `projectId` map, so we can just use it as is.

src/utils/jobState.ts
```diff
@@ -3,18 +3,26 @@ import { cloudresourcemanager_v3, compute_v1 } from 'googleapis';
...
 export const PEERED_NETWORKS = 'network:all_peerings';
 
-export async function cacheProjectNameToId(
+export async function cacheProjectNameAndId(
   jobState: JobState,
   project: cloudresourcemanager_v3.Schema$Project,
 ) {
-  await jobState.setData(`project:${project.name}`, project.projectId);
+  await jobState.setData(`projectName:${project.name}`, project.projectId);
+  await jobState.setData(`projectId:${project.projectId}`, project.name);
```
mknoedel, line 11: Need to support both maps for various reasons. IMHO, we should just be using the projectId as the `_key` instead of the name as it is more unique.

src/steps/cloud-asset/index.ts
```diff
@@ -277,11 +281,20 @@ export async function createBindingToAnyResourceRelationships(
...
-                targetFilterKeys: [['_type', '_key']],
+                targetFilterKeys: [
+                  // Because there is no one-to-one-mapping from Google Resource Kind to J1 Type, only map on the `_key`.
+                  type === MULTIPLE_J1_TYPES_FOR_RESOURCE_KIND
+                    ? ['_key']
+                    : ['_type', '_key'],
+                ],
                 skipTargetCreation: false,
                 targetEntity: {
                   _type: type,
```
aiwilliams, line 300: Note: `_type` cannot be specified in the case that the `targetEntity` is a SQL database.

aiwilliams (changes requested):

austinkelleher (changes requested):

src/utils/jobState.ts
aiwilliams: It would seem worthwhile to move toward using the `projectId`, perhaps in another PR after considering the implications.

src/steps/cloud-asset/__recordings__/createMappedBindingAnyResourceRelationships_3853245014/recording.har
```diff
@@ -2672,6 +2672,249 @@
...
+          "method": "GET",
+          "queryString": [],
+          "url": "https://cloudresourcemanager.googleapis.com/v3/projects/j1-gc-integration-dev-v3"
+        },
+        "response": {
+          "bodySize": 617,
+          "content": {
+            "mimeType": "application/json; charset=UTF-8",
+            "size": 617,
+            "text": "{\"error\":{\"code\":401,\"message\":\"Request had invalid authentication credentials. Expected OAuth 2 access token, login cookie or other valid authentication credential. See https://developers.google.com/identity/sign-in/web/devconsole-project.\",\"errors\":[{\"message\":\"Invalid Credentials\",\"domain\":\"global\",\"reason\":\"authError\",\"location\":\"Authorization\",\"locationType\":\"header\"}],\"status\":\"UNAUTHENTICATED\"}}"
```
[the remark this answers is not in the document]
mknoedel: The credentials we have in the test account don't allow us to fetch projects so when `fetchResourceManagerProject` is called, it fails the API call. However,  `fetchResourceManagerProject` makes a shell project entity even on api failures. I can then use to check mapped relationships against for my test. 

It isn't important that the project is pulled from the api for this test, only that the entity is in the job state, so I think it is ok to leave this recording in here.

austinkelleher (approved):

aiwilliams (approved):

aiwilliams (approved):

PR #280, opened 2021-07-31
INT-1302 binding to sql relationships

mknoedel:
Needed to change the `_key` property from SQL resources to something that can be found from the resource identifiers. Also changed the relationship mappings to only check the `_key` property which should be fine because the key is a url which is globally unique.

### DIFF
--- a/src/steps/cloud-asset/index.ts
+++ b/src/steps/cloud-asset/index.ts
@@ -41,6 +41,7 @@ import {
   makeLogsForTypeAndKeyResponse,
 } from '../../utils/iamBindings/getTypeAndKeyFromResourceIdentifier';
 import { getEnabledServiceNames } from '../enablement';
+import { DO_SOMETHING_DIFFERENT_FOR_SQL_INSTANCES } from '../../utils/iamBindings/resourceKindToTypeMap';
 
 export async function fetchIamBindings(
   context: IntegrationStepContext,
@@ -288,7 +289,11 @@ export async function createBindingToAnyResourceRelationships(
               _mapping: {
                 relationshipDirection: RelationshipDirection.FORWARD,
                 sourceEntityKey: bindingEntity._key,
-                targetFilterKeys: [['_type', '_key']],
+                targetFilterKeys: [
+                  type === DO_SOMETHING_DIFFERENT_FOR_SQL_INSTANCES
+                    ? ['_key']
+                    : ['_type', '_key'],
+                ],
                 skipTargetCreation: false,
                 targetEntity: {
                   _type: type,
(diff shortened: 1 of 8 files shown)


src/steps/cloud-asset/index.ts
```diff
@@ -288,7 +289,11 @@ export async function createBindingToAnyResourceRelationships(
               _mapping: {
                 relationshipDirection: RelationshipDirection.FORWARD,
                 sourceEntityKey: bindingEntity._key,
-                targetFilterKeys: [['_type', '_key']],
+                targetFilterKeys: [
+                  type === DO_SOMETHING_DIFFERENT_FOR_SQL_INSTANCES
+                    ? ['_key']
+                    : ['_type', '_key'],
```
mknoedel, line 295: This could be more clear somehow, perhaps I make a `targetFilterKeys` variable first?  the variable name could be more descriptive as well perhaps

src/utils/iamBindings/typeToKeyGeneratorMap.ts
```diff
@@ -153,7 +153,7 @@ export const J1_TYPE_TO_KEY_GENERATOR_MAP: {
   [ENTITY_TYPE_REDIS_INSTANCE]: customPrefixAndIdKeyMap(getRedisKey),
   [ENTITY_TYPE_MEMCACHE_INSTANCE]: customPrefixAndIdKeyMap(getMemcacheKey),
   [MONITORING_ALERT_POLICY_TYPE]: fullPathKeyMap,
-  [DO_SOMETHING_DIFFERENT_FOR_SQL_INSTANCES]: impossible, // needs access to the region (us-east1, etc...) the instance is in in order to generate the key
+  [DO_SOMETHING_DIFFERENT_FOR_SQL_INSTANCES]: selfLinkKeyMap,
```
austinkelleher, line 156: Can we change this property name? Maybe something like: `ALL_SQL_INSTANCE_ENTITY_TYPES`.

For data like this, we can use `ALL_` as a prefix. Thoughts?

src/steps/cloud-asset/index.ts
```diff
@@ -288,7 +289,11 @@ export async function createBindingToAnyResourceRelationships(
               _mapping: {
                 relationshipDirection: RelationshipDirection.FORWARD,
                 sourceEntityKey: bindingEntity._key,
-                targetFilterKeys: [['_type', '_key']],
+                targetFilterKeys: [
+                  type === DO_SOMETHING_DIFFERENT_FOR_SQL_INSTANCES
```
austinkelleher, line 293: Can you make a comment about the reasoning for this hack? 

If we have more cases like this, we could introduce another map like:

> graph object type -> targetFilterKeys

For now, we should just leave it as a hack until we see a pattern.

austinkelleher (approved):
I approved with a few suggestions. Can address in this PR or a separate PR. Either is fine. Thanks!

austinkelleher (approved):

PR #283, opened 2021-08-02
Release v0.47.2

mknoedel:


.github/workflows/build.yml
```diff
@@ -56,7 +56,7 @@ jobs:
       - name: Fetch tags
         run: |
           git fetch --tags
-          if git describe --exact-match --match "v*.*.*" HEAD^2
+          if git describe --exact-match --match "v*.*.*" HEAD
```
austinkelleher, line 59: Can you explain this change? Looks like the integration template does not use this method: https://github.com/JupiterOne/integration-template/blob/main/.github/workflows/build.yml#L52

PR #285, opened 2021-08-10
Add 429 code to the existing client retry logic

kenanwarren:


ceelias (dismissed):

ndowmon (approved):
Looks good @kenanwarren. Just a note - if either of these errors (`403` or `429`) return something like `retryAfter`, we could further improve this to make sure our retries are respecting any error response arguments.

ceelias (approved):

kenanwarren:
> Looks good @kenanwarren. Just a note - if either of these errors (`403` or `429`) return something like `retryAfter`, we could further improve this to make sure our retries are respecting any error response arguments.

There doesn't look to be anything in the documentation around a retryAfter header the closest documented thing I found was here: https://cloud.google.com/storage/docs/retry-strategy#exponential-backoff
This just documents their recommended exponential backoff strategy for the storage apis.

PR #288, opened 2021-08-13
Make permissions a string as it is too big to store in Neptune as an array

mknoedel:
This is a breaking change for queries on this key. There are no managed questions that use this key.

Users can still query for a specific permission via `find google_iam_role with permission~={{PERMISSION_NAME}}`.

Permissions are already analyzed to create binding ALLOWS resource relationships so I don't think there will be much value in querying for a specific permission on a role.


CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Changed
+
+- property `permissions` on `google_iam_bindings` to be a string instead of an
+  array.
```
aiwilliams, line 14: It's helpful to form these as complete sentences when bringing them into release notes. Additionally, I think it's valuable to consider expressing the impact on users.

```
- Changed `google_iam_bindings.permissions: string[]` to `google_iam_bindings.permissions: string` due to
  limitations on `string[]` property lengths in JupiterOne. Queries will need to be changed from .... to ....
```

aiwilliams (changes requested):

aiwilliams:
> Users can still query for a specific permission via find google_iam_role with permission~={{PERMISSION_NAME}}, but this query will likely be quite slow

The use of the includes matcher, `~=`, is promoted in a number of queries, so I think it's fine to recommend.

CHANGELOG.md
mknoedel: Sounds good to me going forward to put more context into the CHANGELOG. I've kept it brief for most of my changes in the past, but I can definitely see 

> > Users can still query for a specific permission via find google_iam_role with permission~={{PERMISSION_NAME}}, but this query will likely be quite slow
> 
> The use of the includes matcher, `~=`, is promoted in a number of queries, so I think it's fine to recommend.

Works for me, I wasn't aware if there were query speed differences or not.

mknoedel:
> > Users can still query for a specific permission via find google_iam_role with permission~={{PERMISSION_NAME}}, but this query will likely be quite slow
> 
> The use of the includes matcher, `~=`, is promoted in a number of queries, so I think it's fine to recommend.

Works for me, I wasn't aware if there were query speed differences or not.

aiwilliams (approved):
Looks good to me! I think it a good idea to get @austinkelleher's input on this one, as you suggested in Slack.

austinkelleher (approved):

aiwilliams:
Created https://jupiterone.atlassian.net/browse/CORE-1470 to get the persister to let us know.

PR #291, opened 2021-08-16
INT-1560 - Managed roles do not have permissions

mknoedel:
Be sure to add permissions to roles that are created via mapped relationship.


src/steps/iam/index.ts
```diff
@@ -144,10 +144,12 @@ export async function fetchIamManagedRoles(
 ): Promise<void> {
   const { jobState, instance } = context;
   const client = new IamClient({ config: instance.config });
-  const managedRoles: iam_v1.Schema$Role[] = [];
+  const managedRoles: { [k: string]: iam_v1.Schema$Role } = {};
```
austinkelleher, line 147: Nice!

src/utils/iam.ts
```diff
@@ -193,11 +193,14 @@ export async function getIamManagedRoleData(
...
 
-export function buildPermissionsByApiServiceMap(roles: iam_v1.Schema$Role[]) {
+export function buildPermissionsByApiServiceMap(roles: {
+  [k: string]: iam_v1.Schema$Role;
+}) {
   const allPermissions: Set<string> = new Set();
   const permissionsByServiceMap: Map<string, string[]> = new Map();
 
-  for (const role of roles) {
+  for (const roleName of Object.keys(roles)) {
```
austinkelleher, line 202: Minor: Using `Object.entries` would simplify this code slightly I think.

austinkelleher (approved):

PR #293, opened 2021-08-17
[DO NOT MERGE] INT-1561  - add private, publicRead or publicWrite on all resources

mknoedel:
### WHAT WAS DONE
Analyze the organization level `google_iam_bindings` to determine the access level of every resource that has a binding.

image

### POTENTIAL ISSUE WITH THIS SOLUTION
A `google_storage_bucket`'s accessLevel is dependent on other configuration factors only available on the bucket itself that override any IAM policy analysis. Because of this, if a mapped `google_iam_binding_allows_ANY_RESOURCE` relationship links up with a `google_storage_bucket` that does not have `uniformBucketLevelAccess` enabled that otherwise would be considered `private`, it's accessLevel would be updated from `subjectToObjectACLs` to `['subjectToObjectACLs', 'private']` thanks to the mapper. I determined this to be fine as we do not lose the `subjectToObjectACLs` value, though it might cause some confusion in the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,15 @@ and this project adheres to
 
 ### Added
 
+- Added support for ingesting the following **new** properties on direct and
+  mapped `google_iam_binding_allows_ANY_RESOURCE` relationships:
+
+  | Property                   | Type                                   | Description                                           |
+  | -------------------------- | -------------------------------------- | ----------------------------------------------------- |
+  | `accessLevel`              | 'private', 'publicRead', 'publicWrite' | The level public access on the contents of the bucket |
+  | `accessLevelIsConditional` | boolean                                | Whether there is a condition to the access or not     |
+  | `accessLevelCondition`     | string                                 | The condition for access at this access level         |
+
 - Added support for ingesting the following **new** properties on
   `google_storage_bucket`:
 
--- a/src/steps/cloud-asset/index.test.ts
+++ b/src/steps/cloud-asset/index.test.ts
@@ -39,6 +39,7 @@ import {
   CLOUD_STORAGE_BUCKET_ENTITY_TYPE,
 } from '../storage';
 import { CLOUD_FUNCTION_ENTITY_TYPE, fetchCloudFunctions } from '../functions';
+import { AccessData } from '../../utils/jobState';
 
 expect.extend({
   toHaveOnlyDirectRelationships(
@@ -266,6 +267,12 @@ describe('#fetchIamBindings', () => {
         encounteredTypes: context.jobState.encounteredTypes,
       }).toMatchSnapshot();
 
+      const accessLevels: AccessData['accessLevel'][] = [
+        'private',
+        'publicRead',
+        'publicWrite',
+      ];
+
       // Relationships
       const {
         google_iam_binding_uses_role,
@@ -325,8 +332,15 @@ describe('#fetchIamBindings', () => {
       ).toHaveOnlyMappedRelationships(
         'google_iam_binding_allows_cloud_project',
       );
+      google_iam_binding_allows_cloud_project.forEach((mappedRelationship) => {
+        expect(accessLevels).toContain(
+          (mappedRelationship as MappedRelationship)._mapping.targetEntity
+            .accessLevel,
+        );
+      });
 
       // Direct Relationships
+      // principal relationships
       expect(
         google_iam_binding_assigned_service_account,
       ).toHaveOnlyDirectRelationships(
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,15 @@ and this project adheres to
 
 ### Added
 
+- Added support for ingesting the following **new** properties on direct and
+  mapped `google_iam_binding_allows_ANY_RESOURCE` relationships:
+
+  | Property                   | Type                                   | Description                                           |
+  | -------------------------- | -------------------------------------- | ----------------------------------------------------- |
+  | `accessLevel`              | 'private', 'publicRead', 'publicWrite' | The level public access on the contents of the bucket |
+  | `accessLevelIsConditional` | boolean                                | Whether there is a condition to the access or not     |
```
ndowmon, line 19: Is this necessary? I think we could rewrite queries with 

```FIND * AS gcp_resource WITH access_level = 'publicRead' AND accessLevelIsConditional = true```

as just

```FIND * AS gcp_resource WITH access_level = 'publicRead' AND accessLevelCondition != undefined```

File: CHANGELOG.md [hunk 2]
```diff
@@ -10,6 +10,15 @@ and this project adheres to
 
 ### Added
 
+- Added support for ingesting the following **new** properties on direct and
+  mapped `google_iam_binding_allows_ANY_RESOURCE` relationships:
```
ndowmon, line 14: Each resource can have many relationships to `google_iam_binding` entities. How does the target entity behave when one of those bindings is `private`, one is `publicRead` with conditions, and one is `publicWrite` without conditions?

ndowmon: Also, can IAM bindings target an entire project? What happens when a `google_iam_binding` targets a project with `accessLevel='publicRead'`, but a different `google_iam_binding` targets a resource within that project with `accessLevel='private'`? Will this target resource be labeled as `pubilcRead` or `private`?

ndowmon: I think the intention here is to answer a question like this:

```
FIND 
  google_storage_bucket WITH accessLevel != 'private'
```

But what if instead, the `accessLevel` was on the `google_iam_binding` entity? You could instead write a query like the following:

```
FIND 
  google_storage_bucket 
  (THAT HAS google_cloud_project)? 
  THAT ALLOWS google_iam_binding WITH accessLevel != 'private'
```

That could ensure the following behavior:
1. Resources with multiple `google_iam_bindings` don't have an array for `accessLevel` and `accessLevelCondition`
2. These types of queries will work consistently, whether this is from a mapped relationship or direct relationship,
3. `accessLevel` properties for any parent resource (`project`, and possibly also `folder` and `organization`) can also be used and queried

mknoedel: > Each resource can have many relationships to `google_iam_binding` entities. How does the target entity behave when one of those bindings is `private`, one is `publicRead` with conditions, and one is `publicWrite` without conditions?

Because of the way I calculate the accessLevel, each binding will have the same accessLevel that represents the most open that is available. You can look at [this test file](https://github.com/JupiterOne/graph-google-cloud/pull/292/files#diff-f84e2c4e66f8ce513216d97ea03bdacfa2bf6ed817249ecb3a9af119929aa9cbR14) to see how "most open" is determined.

This does bring up a good point though that the relationship properties themselves should be related exclusively to the specific binding instead of the aggregate total of all the bindings. 

- [ ] Relationship Access Level properties should relate only to that specific iam_binding.

mknoedel: > I think the intention here is to answer a question like this:
> 
> ```
> FIND 
>   google_storage_bucket WITH accessLevel != 'private'
> ```
> 
> But what if instead, the `accessLevel` was on the `google_iam_binding` entity? You could instead write a query like the following:
> 
> ```
> FIND 
>   google_storage_bucket 
>   (THAT HAS google_cloud_project)? 
>   THAT ALLOWS google_iam_binding WITH accessLevel != 'private'
> ```
> 
> That could ensure the following behavior:
> 
> 1. Resources with multiple `google_iam_bindings` don't have an array for `accessLevel` and `accessLevelCondition`
> 2. These types of queries will work consistently, whether this is from a mapped relationship or direct relationship,
> 3. `accessLevel` properties for any parent resource (`project`, and possibly also `folder` and `organization`) can also be used and queried

This makes a ton of sense to me. It also prevents the need to need to calculate if a resource is public at the ingestion step level which causes a good amount of confusion.

mknoedel: > Also, can IAM bindings target an entire project? What happens when a `google_iam_binding` targets a project with `accessLevel='publicRead'`, but a different `google_iam_binding` targets a resource within that project with `accessLevel='private'`? Will this target resource be labeled as `pubilcRead` or `private`?

Per https://cloud.google.com/iam/docs/resource-hierarchy-access-control, there is a structured hierarchy for access control. However, for the purposes of determining if a resource is 'private', 'publicRead' or 'publicWrite', the only thing that is considered is if there is a binding with `allUsers` or `allAuthenticatedUsers` or not. Per https://cloud.google.com/resource-manager/reference/rest/v1/projects/setIamPolicy (and by using the console), there isn't a way to create a binding at the organization, folder, or project level to either `allUsers` or `allAuthenticatedUsers`. Because of this, for the purpose of AccessLevel, we just need to make sure the hierarchy in the cloud services themselves are properly represented.

CHANGELOG.md
mknoedel: That's true, I'll go ahead and remove it.

CHANGELOG.md
mknoedel: I think it makes sense to put the accessLevel on the binding makes a lot of sense. It does put more pressure on the query side as opposed to the ingestion side, but for that sacrifice it makes it so we don't have to rely on the mapper as much and saves a lot of strange jobState storage stuff.

PR #296, opened 2021-08-19
INT-1561 - Add isPublic and isOpenToTheInternet to IAM bindings

mknoedel:
### Added

- New properties added to resources:

  | Entity                | Properties            |
  | --------------------- | --------------------- |
  | `google_iam_bindings` | `isReadOnly`          |
  | `google_iam_bindings` | `isOpenToTheInternet` |

#### Why Put Access Info on the Binding instead of the Resource?

Per [@ndowmon's suggestion in the previous INT-1561 PR](https://github.com/JupiterOne/graph-google-cloud/pull/293/files#r691460798), instead of allowing users to find the access level of various resources via querying that resource directly. Example:

```
FIND
  google_storage_bucket WITH accessLevel != 'private'
```

We can instead put the access information on the IAM binding and have users do a traversal query. Example:

```
FIND 
  google_storage_bucket 
  THAT ALLOWS google_iam_binding WITH accessLevel != 'private'
```

From the users perspective, this is slightly worse as they are expected to know the data structure of Role Bindings to Resources in order to form the query. However, from the ingestion perspective, this is far easier because we do not need to enhance entities with information, which our sdk does not support yet (See [sdk issue #532](https://github.com/JupiterOne/sdk/issues/532)).

If we needed to support the first query today, we would need to restructure the graph project in such a way where all access information is available at the time of creating entities, which I did in these 2 PRs ([INT-1142](https://github.com/JupiterOne/graph-google-cloud/pull/292/files#diff-575c3f04741892b0dd274bf9163166e8b560b5febde394fbb61344d572b69ac3), [INT-1561](https://github.com/JupiterOne/graph-google-cloud/pull/293/files)). Unfortunately, doing so makes this project significantly more difficult to work with in the future. Take a look for yourself, it's a bit hard to follow.

#### Why use `isReadOnly` and `isOpenToTheInternet` instead of an `accessLevel` of 'private', 'publicRead 'and 'publicWrite' values?
It felt to me that it would more clear to users making queries to put things into booleans instead of needing to have the user guess what the possible enum values of an entity property could be. How would a user know that the available accessLevels are `private` | `publicRead` | `publicWrite`? From what I can tell, the only ways would be we tell them or they look at the code. This problem goes away when you use boolean properties because each property is returned in J1QL queries.  My guess is that people can guess the available values of `isReadOnly` and `isOpenToTheInternet` quite easily. Of course another way to solve this problem would be to provide type docs for each entity, but we do not have those yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
+- New properties added to resources:
+
+  | Entity                | Properties            |
+  | --------------------- | --------------------- |
+  | `google_iam_bindings` | `isReadOnly`          |
+  | `google_iam_bindings` | `isOpenToTheInternet` |
+
 ### Fixed
 
 - Managed `google_iam_roles` now have a `permissions` property, similar to
--- a/docs/jupiterone.md
+++ b/docs/jupiterone.md
@@ -265,7 +265,7 @@ NOTE: ALL OF THE FOLLOWING DOCUMENTATION IS GENERATED USING THE
 "j1-integration document" COMMAND. DO NOT EDIT BY HAND! PLEASE SEE THE DEVELOPER
 DOCUMENTATION FOR USAGE INFORMATION:
 
-https://github.com/JupiterOne/sdk/blob/master/docs/integrations/development.md
+https://github.com/JupiterOne/sdk/blob/main/docs/integrations/development.md
 ********************************************************************************
 -->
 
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,15 @@ and this project adheres to
...
 ## [Unreleased]
 
+### Added
+
+- New properties added to resources:
+
+  | Entity                | Properties            |
+  | --------------------- | --------------------- |
+  | `google_iam_bindings` | `isReadOnly`          |
+  | `google_iam_bindings` | `isOpenToTheInternet` |
```
aiwilliams, line 18: Is this a term that users will be familiar with, saying a binding is "open to the Internet"? Why not "isMemberPublic" like the function?

PR #304, opened 2021-08-27
INT-1669 missing project

mknoedel:
Fetching bindings, we were sometimes getting the error `Missing project ID in local cache`. After some investigation, I found that it was because bindings were referencing projects that were deleted. The fix was to query for projects that were recently deleted. This has the benefits of 1. being able to map all project `google_iam_bindings` and 2. be able to set the project.state to "DELETED" as soon as possible.


austinkelleher (approved):
Nice change!

PR #305, opened 2021-08-30
INT-1671 - Never use entity creation in mapped relationships as it does not work right now

mknoedel:


austinkelleher (approved):
I think a remap will be required to remove the target entities that were created

PR #313, opened 2021-09-10
Make relationships between project and storage bucket in order to do proper IAM analysis

mknoedel:
Right now, we are only able to do analysis of the roles attached directly to Google Cloud resources. We need to connect those resources to the projects so we can analyze the roles on the project, folder and org levels as well.

This is just the first, but we will need every resource that can be the target of a role binding to have a relationship with the project.


mknoedel:
We no longer want to make relationships directly from the bucket to the project. We will go through the service first now.

PR #322, opened 2021-09-24
INTS-23: Fix missing Internet --ALLOWS--> Firewall relationship

rwaterman:
Fixes missing internet allows firewall relationship from appearing in the graph on entities with `ingress: true`  + and an IP range of `0.0.0.0/0` by: setting `targetFilterKeys` to use both `_key` *and* `_type` (see existing `isInternet` usage for more info).

I'm trying to come up with a better explanation of why this worked so if anyone would like to enlighten me, please feel free to discuss more.

**Steps to Reproduce**
1. Create a new firewall entity that has not been used before (`gcloud compute firewall-rules create some-new-name --allow=tcp:XXXX --source-ranges="0.0.0.0/0" --description="Allow incoming traffic" --direction=INGRESS`)
2. Run your custom integration from the  `main` branch (`yarn run -i <id> -d`)
3. Observe the firewall has no internet allows connection. You may want to wait a few minutes to verify all sync tasks have completed.
4. Feel free to create another new firewall entity if desired (optional)
5. Deploy changes from this branch with the fix
6. Observe that internet connection is set correctly (the synchronization may take about 5 minutes to finish)

If you deploy back to main without this fix (maybe you want to test this again), you'll want to use a new firewall name that will have new relationship keys as the old relationship keys added in the fix may not get deleted automatically (i.e., if the integration is not tagged).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ dist/
 tsconfig.tsbuildinfo
 coverage/
 main.remote.tf
+.DS_Store
+.idea/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Fixed
+
+- `Internet` **ALLOWS** `google_compute_firewall` relationship for
+  `0.0.0.0/0`/`::/0` source CIDR blocks restored by adding `_type` to the target
+  filter keys.
+
 ## 0.50.0 - 2021-09-15
 
 ### Added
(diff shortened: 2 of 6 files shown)


ceelias (dismissed):
Nice! Generally like to include the Changlog changes in this PR since its more of a pain to makes changes to it when you add it in the Version PR since it's tagged. No big deal at all though.

austinkelleher (dismissed):

ceelias (approved):

rwaterman (commented):
Discussed running some J1Q1 during testing to avoid any issues that *may* be relevant from  https://github.com/JupiterOne/graph-google-cloud/pull/322. If any queries you decide to run to ensure relationships look sane in the UI are good, I'm good.

PR #330, opened 2021-09-30
Prevent unable to find project id error

mknoedel:
Not every Google Cloud organization structure will have access to the projectId in the jobState, so instead of erroring, we just store the projectName instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "format": "prettier --write '**/*.{ts,js,json,css,md,yml}'",
     "type-check": "tsc",
     "test:env": "LOAD_ENV=1 yarn test",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -23,7 +23,7 @@
     "format": "prettier --write '**/*.{ts,js,json,css,md,yml}'",
     "type-check": "tsc",
     "test:env": "LOAD_ENV=1 yarn test",
-    "test": "jest && yarn validate:questions:dry",
```
mknoedel, line 26: This is so we can do things like `yarn test -u`

austinkelleher (approved):

PR #331, opened 2021-09-30
Add domainName on google user entities

mknoedel:
Needed to make domain -> user relationships which is a good relationship for Google Cloud IAM access parsing

### DIFF
--- a/src/steps/resource-manager/createIamEntities.ts
+++ b/src/steps/resource-manager/createIamEntities.ts
@@ -12,6 +12,7 @@ import {
   PrimitiveEntity,
 } from '@jupiterone/integration-sdk-core';
 import { IamUserEntityWithParsedMember } from '.';
+import { last } from 'lodash';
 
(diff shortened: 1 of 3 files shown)


src/steps/resource-manager/createIamEntities.ts
```diff
@@ -22,6 +22,7 @@ import {
   PrimitiveEntity,
 } from '@jupiterone/integration-sdk-core';
 import { IamUserEntityWithParsedMember } from '.';
+import { last } from 'lodash';
```
austinkelleher, line 25: Same feedback here: https://github.com/JupiterOne/graph-google/pull/114#discussion_r725135306

This doesn't seem to be a direct dependency. Additionally, lodash is a pretty huge package to bundle in for this function.

How about a util like this:

```ts
function last<T>(arr: T[] | undefined): T {
  if (!arr || !arr.length) return undefined;
  return arr[arr.length - 1];
}

function getDomainFromGoogleUser(
  data: admin_directory_v1.Schema$User
): string | undefined {
  return last(data.primaryEmail?.split('@'));
}
```

PR #333, opened 2021-10-01
Create Api Service HAS Any Resource relationships

mknoedel:
Right now, we are only able to do analysis of the roles attached directly to Google Cloud resources. We need to connect those resources to the projects so we can analyze the roles on the project, folder and org levels as well.

#### Why not connect resources to their projects directly instead of going through the Api Service first?

In order to facilitate a better user experience with the front-end node expansion functionality:
image
... we don't want to have too many things attached to the project as it would get unruley very fast. It is a bit cleaner to attach it to the api service instead. Plus, that is what Google Cloud does anyway according to their Resource Hierarchy graph:
image


src/steps/cloud-asset/index.ts
```diff
@@ -399,6 +405,110 @@ export async function createBindingToAnyResourceRelationships(
...
+  // Optimization to prevent calling jobState.findEntity for entities we have already made relationships for.
+  const projectToResourceTracker = new Set<string>();
+
+  await jobState.iterateEntities(
+    { _type: bindingEntities.BINDINGS._type },
+    async (bindingEntity: BindingEntity) => {
+      // skip processing resource -> project pairs that we have already made relationships for.
+      if (
+        projectToResourceTracker.has(
+          (bindingEntity.projectId ?? bindingEntity.projectName) +
```
austinkelleher, line 422: Let's create a function to help clean this up a bit. You can also remove the `else` and re-use a calculated resource key.

Consider something like:

```typescript
const resourceKey = buildBindingEntityResourceKey(bindingEntity);

if (projectToResourceTracker.has(resourceKey)) return;
projectToResourceTracker.add(resourceKey);
```

src/steps/cloud-asset/index.ts
```diff
@@ -399,6 +405,110 @@ export async function createBindingToAnyResourceRelationships(
   );
 }
 
+export async function createApiServiceToAnyResourceRelationships(
```
austinkelleher, line 408: Overall I think the logic in this function looks good, but there is a lot going on. Can we split this up into a few smaller functions?

src/steps/cloud-asset/index.ts
```diff
@@ -399,6 +405,110 @@ export async function createBindingToAnyResourceRelationships(
...
+      ) ?? {};
+
+      // Do NOT make relationships for resources in the Organization Resource Heirarchy
+      if (
+        [
+          ORGANIZATION_ENTITY_TYPE,
+          FOLDER_ENTITY_TYPE,
+          PROJECT_ENTITY_TYPE,
+          API_SERVICE_ENTITY_TYPE,
+        ].includes(resourceType!)
```
austinkelleher, line 455:
```ts
if (isOrganizationalHierarchyResource(resourceType)) {

}
```

src/steps/cloud-asset/index.ts
```diff
@@ -399,6 +405,110 @@ export async function createBindingToAnyResourceRelationships(
...
+      ) {
+        return;
+      }
+
+      if (googleResourceKind && resourceKey) {
+        const serviceKey =
```
austinkelleher, line 461:
```ts
const serviceGraphObjectKey = buildServiceGraphObjectKeyFromBindingEntity(bindingEntity);
```

src/steps/cloud-asset/index.ts
```diff
@@ -399,6 +405,110 @@ export async function createBindingToAnyResourceRelationships(
...
+          (bindingEntity.projectId ?? bindingEntity.projectName) +
+          '/services/' +
+          googleResourceKind.split('/')[0];
+        const serviceEntity = await jobState.findEntity(serviceKey);
+        const resourceEntity = await jobState.findEntity(resourceKey);
+
+        // This means we can not make these relationships for sqladmin.googleapis.com/Instances
+        if (
+          !serviceEntity ||
+          resourceType === MULTIPLE_J1_TYPES_FOR_RESOURCE_KIND ||
```
austinkelleher, line 472: I remember this being an issue, but I'm a bit fuzzy on the conclusion of whether this is permanent. Can you elaborate on whether this is permanent or not? If it's not permanent, can you explain how we can get rid of this?

src/steps/cloud-asset/index.ts
```diff
@@ -399,6 +405,110 @@ export async function createBindingToAnyResourceRelationships(
...
+          !resourceType
+        )
+          return;
+        await jobState.addRelationship(
+          resourceEntity
+            ? createDirectRelationship({
```
austinkelleher, line 478: We seem to have this type of conditional logic inside of the project a lot. It would be nice if we introduced a `createDirectOrMappedRelationship` function. Doesn't have to happen in this PR, but it would help clean up code like this.

Example:

```ts
await jobState.addRelationship(
  createDirectOrMappedRelationship({
  
  })
);
```

src/steps/cloud-asset/index.ts
```diff
@@ -399,6 +405,110 @@ export async function createBindingToAnyResourceRelationships(
...
+            : createMappedRelationship({
+                _class: RelationshipClass.HAS,
+                _type: generateRelationshipType(
+                  RelationshipClass.HAS,
+                  API_SERVICE_ENTITY_TYPE,
+                  resourceType,
+                ),
+                source: serviceEntity,
+                /**
+                 * The mapper does properly remove mapper-created entities at the moment. These
```
austinkelleher, line 492: `"does not properly"`?

austinkelleher (changes requested):
Looks good! Left some suggestions.

src/steps/cloud-asset/index.ts
mknoedel: The issue here is the way that we modeled sql instances is not 1 to 1 with our `_type`s. A Google Resource Type of `sqladmin.googleapis.com/Instances` could be `google_sql_mysql_instance`, `google_sql_postgres_instance`, or `google_sql_sql_server_instance` and I'm not sure of any way to know which one it is. 

In order to fix this, we will have to change these types.

src/steps/cloud-asset/index.ts
mknoedel: Totally agree. I think this makes sense as a cleanup effort in another PR as it would likely want to use this pattern in many places and I'm worried it would muddle this PR.

mknoedel: I tried making this method, but it ended up seeming a bit more cumbersome than helpful. I still need to put the conditional logic somewhere in order to generate the targetEntity properly for the mapped relationship. I think the reason why this Direct or Mapped logic looks confusing is because of the large comment over skipTargetCreation.

austinkelleher (approved):

PR #334, opened 2021-10-01
Traverse the whole resource hierarchy when analyzing convenience members

mknoedel:
TODO: document

### DIFF
--- a/src/steps/cloud-asset/index.ts
+++ b/src/steps/cloud-asset/index.ts
@@ -4,7 +4,6 @@ import {
   generateRelationshipType,
   getRawData,
   IntegrationStep,
-  Relationship,
   RelationshipClass,
   RelationshipDirection,
 } from '@jupiterone/integration-sdk-core';
@@ -18,6 +17,7 @@ import {
   buildIamTargetRelationship,
   findOrCreateIamRoleEntity,
   getPermissionsForManagedRole,
+  isConvienenceMember,
   maybeFindIamUserEntityWithParsedMember,
 } from '../resource-manager';
 import { CloudAssetClient } from './client';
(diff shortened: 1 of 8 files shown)


src/steps/resource-manager/index.ts
```diff
@@ -84,25 +80,15 @@ export async function maybeFindIamUserEntityWithParsedMember({
...
-    // find the Basic Role that relates to this member.
-    const [convenienceMember, projectId] = member.split(':');
-    const projectKey =
-      (await getProjectNameFromId(context!.jobState, projectId)) ?? projectId;
-    const roleKey =
-      projectKey +
-      '/' +
-      getRoleKeyFromConvienenceType(convenienceMember as ConvenienceMemberType);
-    // TODO: need to check for all Folder and Organization keys as well.
-    userEntity = await jobState.findEntity(roleKey);
```
mknoedel, line 97: Convenience Members will no longer be handled in this method

src/steps/cloud-asset/index.ts
```diff
@@ -64,7 +71,8 @@ export async function fetchIamBindings(
   try {
     await client.iterateAllIamPolicies(context, async (policyResult) => {
       const resource = policyResult.resource;
-      const projectName = policyResult.project as string | undefined;
+      const projectName = policyResult.project;
+      const { organization, folders } = policyResult as any;
```
mknoedel, line 75: These are on the policy response even though the SDK says they are not.

src/steps/resource-manager/index.ts
mknoedel: In the next PR when I clean up `fetchResourceManagerIamPolicy`, this method will be removed entirely.

src/steps/cloud-asset/index.ts
```diff
@@ -276,49 +292,80 @@ export async function createBindingRoleRelationships(
...
-          buildIamTargetRelationship({
+            bindingEntity,
+            condition,
+          );
+        } else {
+          const iamUserEntityWithParsedMember =
+            await maybeFindIamUserEntityWithParsedMember({
+              context,
+              member,
+            });
```
mknoedel, line 359: It's kind of weird to have a split between convenience members and other members, so in the next PR when I clean up fetchResourceManagerIamPolicy, this method will be removed entirely.

src/steps/cloud-asset/index.ts
```diff
@@ -467,9 +514,10 @@ export async function createApiServiceToAnyResourceRelationships(
       if (isOrganizationalHierarchyResource(resourceType!)) return;
       if (!googleResourceKind || !resourceKey) return;
 
+      const serviceIdentifier = googleResourceKind.split('/')[0]; // 'cloudfunctions.googleapis.com/CloudFunction' => 'cloudfunctions.googleapis.com'
```
austinkelleher, line 517: Thanks for providing an example in the code

austinkelleher (approved):

PR #335, opened 2021-10-05
Remove fetchProjectIamPolicy

mknoedel:

PR #341, opened 2021-10-12
Revert deprecating step fetchProjectIamPolicy until 2022

mknoedel:
### What Was Done
- Reverted the removal of `fetchProjectIamPolicy` via #335
- Added `deprecationDate` to all relationships in the stepMetadata for `fetchProjectIamPolicy`.


src/steps/cloud-asset/index.test.ts
```diff
@@ -13,7 +13,7 @@ import {
   createApiServiceToAnyResourceRelationships,
   fetchIamBindings,
 } from '.';
-import { bindingEntities } from './constants';
```
mknoedel, line 16: This was causing a module loading issue in tests.

src/steps/resource-manager/converters.ts
```diff
@@ -83,3 +95,45 @@ export function createProjectEntity(
...
+    },
+    properties: {
+      _type: generateRelationshipType(
+        RelationshipClass.ASSIGNED,
+        targetEntity._type!,
+        roleEntity._type,
+      ),
+      projectId,
+      ...(condition && getConditionRelationshipProperties(condition)),
+      _deprecationDate: '01-01-2022',
```
mknoedel, line 136: New property `_deprecationDate`

mknoedel:
Erkang chose to do an immediate transition instead of a slow rollout. That means that this PR is unnecessary.

PR #386, opened 2021-11-04
Use the Service Account Project for checking the API enablement for enabling IAM Binding and Role steps

mknoedel:
In order to create an JupiterOne integration with Google Cloud, you need to make a service account under a Google Cloud project and give that service account permissions to access APIs in that Google Cloud environment.

When the config option "Configure Organization Projects" is used, that service account is copied to all other automatically generated Google Cloud Project integration instances for them to use as authentication. That service account however, still lives inside the first project's Google Cloud Project.

Because of this, when determining what steps can be enabled for an integration run of these automatically generated Google Cloud Project integration instance, the service account has access to the combination of the enabled APIs for the project it is currently ingesting data for, as well as the enabled APIs of the project the service account lives in.

Historically, we have only ingested data that was enabled in both the integration's project and the service account's project. However, with the `google_iam_binding` change where we are now ingesting bindings in each project instance instead of all in the one organization instance, most users have yet to turn on read only APIs like the Cloud Asset API in the individual Google Cloud projects. They only have them enabled in their main project.

In this PR, in order to allow users to still have `google_iam_binding`s ingested for their accounts without needing them to individually enable the Cloud Asset and IAM APIs for each project, we just use the enablement of the Organization account for Cloud Asset and IAM steps.

This approach does have the limitation where users will not be able to enable or disable Binding ingestion on a per-project basis, they will only be able to enable all-or-nothing binding ingestion at the Organization level.


src/getStepStartStateExtraTests.test.ts
```diff
@@ -0,0 +1,74 @@
...
+  STEP_IAM_BINDINGS,
+  STEP_CREATE_BASIC_ROLES,
+  STEP_CREATE_BINDING_ANY_RESOURCE_RELATIONSHIPS,
+  STEP_CREATE_API_SERVICE_ANY_RESOURCE_RELATIONSHIPS,
+} from './steps/cloud-asset/constants';
+import { STEP_IAM_CUSTOM_ROLES, STEP_IAM_MANAGED_ROLES } from './steps/iam';
+
+/**
+ * NOTE: This test needs to be seperated from the other getStepStartState tests it needs to run with another project context.
+ */
```
mknoedel, line 19: Pulling this test into another file fixed the issue for me, but the testing system in this project needs a rehaul as it is really hard to work with at the moment.

src/__recordings__/getStepStartStatesExtraTests_2281574594/recording.har
```diff
@@ -0,0 +1,746 @@
+{
```
mknoedel, line 1: Needed to remake recordings for this change.

src/getStepStartStates.ts
```diff
@@ -216,8 +216,11 @@ export default async function getStepStartStates(
   const organizationSteps = { disabled: !masterOrgInstance }; // Only run organization steps if you are the master organization.
 
   let enabledServiceNames: string[];
+  let serviceAccountProjectEnabledServiceNames: string[];
   try {
     enabledServiceNames = await enablement.getEnabledServiceNames(config);
+    serviceAccountProjectEnabledServiceNames =
+      await enablement.getEnabledServiceNames(config, true);
```
mknoedel, line 223: There is definitely a better way to do this, but this was the lightest-touch approach I could think of.

Doing this calls the Service Enablement API a second time unnecessarily, but it only happens once per integration run which seemed fine to me. I was cacheing the api results but it was breaking tests in hard to exact ways. 

Let me know if you would like me to update this to be a bit cleaner, or if it is ok in this state.

getStepStartStates should be cleaned up as well.

src/getStepStartStates.ts
```diff
@@ -229,23 +232,40 @@ export default async function getStepStartStates(
...
-    return enablement.createStepStartState(
-      enabledServiceNames,
+  ): StepStartState {
+    return enablement.createStepStartStateWhereAllServicesMustBeEnabled(
+      serviceAccountProjectEnabledServiceNames, // using mainProjectEnabledServiceNames instead of enabledServiceNames
       primaryServiceName,
       ...additionalServiceNames,
     );
-  };
+  }
```
mknoedel, line 262: This is the hack to use the API enablement of the service account project instead of the project getting ingested.

src/getStepStartStates.ts
```diff
@@ -229,23 +232,40 @@ export default async function getStepStartStates(
...
 
-  const createStepStartStateWhereAllServicesMustBeEnabled = (
+  const createStepStartState = (
```
mknoedel, line 264: There is no change here even though Github is making it look like there is. I just deleted the `createStepStartStateWhereAllServicesMustBeEnabled` method as it was no longer getting used which made GitHub's formatting look weird.

src/getStepStartStates.ts
```diff
@@ -229,23 +232,40 @@ export default async function getStepStartStates(
...
     primaryServiceName: ServiceUsageName,
     ...additionalServiceNames: ServiceUsageName[]
   ): StepStartState => {
-    return enablement.createStepStartStateWhereAllServicesMustBeEnabled(
+    return enablement.createStepStartState(
```
mknoedel, line 268: Same here, no actual change. See above.

src/getStepStartStates.ts
```diff
@@ -280,32 +300,31 @@ export default async function getStepStartStates(
...
+    ),
+    [STEP_CREATE_BASIC_ROLES]: createStartStatesBasedOnServiceAccountProject(
       ServiceUsageName.CLOUD_ASSET,
       ServiceUsageName.IAM,
     ),
-    [STEP_CREATE_BASIC_ROLES]:
-      createStepStartStateWhereAllServicesMustBeEnabled(
-        ServiceUsageName.CLOUD_ASSET,
-        ServiceUsageName.IAM,
-      ),
```
mknoedel, line 291: This step was not deleted, just strange formatting by Github

src/getStepStartStates.ts
```diff
@@ -320,8 +339,12 @@ export default async function getStepStartStates(
...
       ServiceUsageName.STORAGE_API,
     ),
-    [STEP_IAM_CUSTOM_ROLES]: createStepStartState(ServiceUsageName.IAM),
-    [STEP_IAM_MANAGED_ROLES]: createStepStartState(ServiceUsageName.IAM),
+    [STEP_IAM_CUSTOM_ROLES]: createStartStatesBasedOnServiceAccountProject(
+      ServiceUsageName.IAM,
+    ),
+    [STEP_IAM_MANAGED_ROLES]: createStartStatesBasedOnServiceAccountProject(
+      ServiceUsageName.IAM,
+    ),
```
mknoedel, line 347: This is an important change. I needed to have managed and custom roles be ingested in order to get the permissions denormalized onto Bindings and to make Basic Roles. This does mean that we are adding potentially more ingestion than we have previously.

I don't think this should be a problem at all as that is what we do all the time. Just something to note.

src/index.test.ts
```diff
@@ -930,7 +930,9 @@ describe('#beforeAddEntity', () => {
         invocationConfig.beforeAddEntity(context, mockEntity),
     ).toEqual({
       ...mockEntity,
-      projectId: 'j1-gc-integration-dev-v2',
+      projectId:
+        context.instance.config.projectId ??
+        context.instance.config.serviceAccountKeyConfig.project_id,
```
mknoedel, line 935: Something really weird was happening with these recordings and tests where if was flipping back and forth between 'j1-gc-integration-dev-v2' and 'j1-gc-integration-dev-v3' when I was running `yarn test` and `yarn test:env`.  This update works and is in line with the test's purpose.

src/index.test.ts
```diff
@@ -195,6 +195,177 @@ async function validateInvocationInvalidConfigTest({
   expect(failed).toEqual(true);
 }
 
+describe('#getStepStartStates failures', () => {
```
mknoedel, line 198: This was just moving the failures above the successes. No new tests here.

src/steps/enablement.test.ts
```diff
@@ -3,25 +3,6 @@ import { ServiceUsageName } from '../google-cloud/types';
 import * as enablement from './enablement';
 import * as serviceUsage from './service-usage/client';
 
-describe('#getUniqueIntegrationConfigProjectsForStepEnablement', () => {
```
mknoedel, line 6: This method is no longer used.

src/steps/enablement.test.ts
```diff
@@ -163,9 +144,10 @@ describe('#getEnabledServiceNames', () => {
 
     const collectEnabledServicesForProjectSpy = jest
       .spyOn(serviceUsage, 'collectEnabledServicesForProject')
-      .mockResolvedValue(Promise.resolve(mainMockEnabledServiceNames))
-      .mockResolvedValue(Promise.resolve(targetMockEnabledServiceNames));
```
mknoedel, line 167: `mockResolvedValue` is an absolute override so `collectEnabledServicesForProject` was just returning `targetMockEnabledServiceNames` twice 🙃

src/steps/enablement.ts
```diff
@@ -3,32 +3,6 @@ import { ServiceUsageName } from '../google-cloud/types';
...
 
-/**
- * An integration can provide a service account key file for a service account
- * that exists in one project, but has permissions to ingest data from a
- * _different_ project. This function determines which unique projects should
- * be considered when calculating step enablement.
- *
- * The first index in the array is guaranteed to be the "main" project
- */
-export function getUniqueIntegrationConfigProjectsForStepEnablement(
```
mknoedel, line 14: no need for this method anymore because of the way we are cacheing the service account projects enablements.

src/steps/enablement.ts
```diff
@@ -83,6 +55,22 @@ export async function getEnabledServiceNames(
   return enabledServicesIntersection;
 }
 
+// Cahceing this value so we don't have to fetch it twice
+let mainProjectEnabledServices: string[] | undefined = undefined;
+export function clearMainProjectEnabledServicesCache() {
+  mainProjectEnabledServices = undefined;
+}
```
mknoedel, line 62: I couldn't find a cleaner way to do this. Let me know if you have suggestions.

austinkelleher (approved):

rwaterman (approved):

PR #392, opened 2021-12-06
INTS-121 - Properly default projectId to serviceAccount projectId when not specified

mknoedel:
This is what is specified in the integration config page, so we should be doing the same in the code.
image

### DIFF
--- a/src/steps/cloud-asset/index.ts
+++ b/src/steps/cloud-asset/index.ts
@@ -147,6 +147,7 @@ export async function fetchIamBindings(
       if (permissions?.length ?? 0 > maxPermissionLength) {
         maxPermissionLength = permissions?.length ?? 0;
       }
+
       const conditionLength =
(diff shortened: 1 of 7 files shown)


austinkelleher (dismissed):

src/steps/cloud-asset/index.ts
```diff
@@ -174,9 +174,9 @@ export async function fetchIamBindings(
   // Need to get bindings for every level of the Resource Hierarchy.
   try {
     // Project level bindings and all resource level bindings in that project
-    if (context.instance.config.projectId) {
+    if (client.projectId) {
       await client.iterateIamPoliciesForProjectAndResources(
-        context.instance.config.projectId,
+        client.projectId,
```
austinkelleher, line 179: Would be great if we could get a test for this to prevent regression!

mknoedel: Added a few tests

austinkelleher (approved):

PR #394, opened 2021-12-06
INTS-122 - Do not store projectId on entities that are not related to the project

mknoedel:
This has caused confusion when using projectId to search off of. A more permeant solution for this might be to go into every converter and manually add the projectId. I opted not to do that in this PR in order to maintain a light-touch, but it would be more clear if we switched to that method.

### DIFF
--- a/src/steps/compute/__snapshots__/index.test.ts.snap
+++ b/src/steps/compute/__snapshots__/index.test.ts.snap
@@ -16621,7 +16621,7 @@ Object {
       "_class": Array [
         "Image",
       ],
-      "_key": undefined,
+      "_key": "j1-gc-integration-dev-v3:example-snapshot-image:deleted",
       "_rawData": Array [
         Object {
           "name": "default",
@@ -17209,8 +17209,8 @@ Object {
     Object {
       "_class": "USES",
       "_fromEntityKey": "disk:3169341475684659331",
-      "_key": "disk:3169341475684659331|uses|undefined",
-      "_toEntityKey": undefined,
+      "_key": "disk:3169341475684659331|uses|j1-gc-integration-dev-v3:example-snapshot-image:deleted",
+      "_toEntityKey": "j1-gc-integration-dev-v3:example-snapshot-image:deleted",
       "_type": "google_compute_disk_uses_image",
       "displayName": "USES",
     },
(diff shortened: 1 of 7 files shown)


src/steps/compute/__snapshots__/index.test.ts.snap
```diff
@@ -16621,7 +16621,7 @@ Object {
       "_class": Array [
```
mknoedel, line 16621: This was unrelated to my change. Previous engineer must have missed a test.

src/utils/maybeDefaultProjectIdOnEntity.test.ts
```diff
@@ -0,0 +1,100 @@
...
+} from '../steps/resource-manager';
+import { maybeDefaultProjectIdOnEntity } from './maybeDefaultProjectIdOnEntity';
+
+describe('maybeDefaultProjectIdOnEntity', () => {
+  [
+    ORGANIZATION_ENTITY_TYPE,
+    FOLDER_ENTITY_TYPE,
+    bindingEntities.BINDINGS._type,
+  ].forEach((entityType) => {
+    it(`should not default a projectId on ${entityType} entities`, () => {
```
austinkelleher, line 18: Minor, but for awareness, we can use `test.each` for this type of thing.

austinkelleher: Btw was there a reason why you chose to use `it`? I don't think we use that much in this project. Would be great to be consistent. We should consider adding a linting rule that enforces one or the other in the future.

austinkelleher (approved):

src/utils/maybeDefaultProjectIdOnEntity.test.ts
mknoedel: I think `it should` reads better than `test should`, but I don't really care either way. I'll switch it.

mknoedel: I'm not actually a fan of `test.each` as the interface is fairly ugly as you need to interpolate with `%s`. imho, `forEach` is a builtin javascript array method that everyone should already intuitively understand. I'll switch it though for consistency sake.

austinkelleher (approved):

PR #396, opened 2021-12-06
Store rawData on google_iam_bindings again

mknoedel:
This was originally taken out because it was assumed that this was causing a RequestTooLarge exception. This was not the issue and can thus be safely be re-added. Plus, the rawData for bindings is actually really small. Example:

```
      "_rawData": [
        {
          "name": "default",
          "rawData": {
            "role": "roles/cloudkms.serviceAgent",
            "members": [
              "serviceAccount:service-167984947943@gcp-sa-cloudkms.iam.gserviceaccount.com"
            ]
          }
        }
      ]
```

### DIFF
--- a/src/steps/compute/__snapshots__/index.test.ts.snap
+++ b/src/steps/compute/__snapshots__/index.test.ts.snap
@@ -16621,7 +16621,7 @@ Object {
       "_class": Array [
         "Image",
       ],
-      "_key": undefined,
+      "_key": "j1-gc-integration-dev-v3:example-snapshot-image:deleted",
       "_rawData": Array [
         Object {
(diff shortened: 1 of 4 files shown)


src/steps/compute/__snapshots__/index.test.ts.snap
```diff
@@ -16621,7 +16621,7 @@ Object {
       "_class": Array [
```
mknoedel, line 16621: unrelated change. Needed to update snapshot to get tests to pass.

austinkelleher (approved):

austinkelleher (approved):

PR #418, opened 2022-01-28
Adding in optional FOLDER_ID to filter by folder

adam-in-ict:
Adding in optional FOLDER_ID to filter down to a specific folder within an organization for ingestion.  This is currently only restricting what folders and projects are ingested and is not doing anything to further restrict other entities.  We may want to rework or add additional filtering based on user or developer feedback.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -51,6 +51,9 @@ export const invocationConfig: IntegrationInvocationConfig<IntegrationConfig> =
         type: 'boolean',
         mask: false,
       },
+      folderId: {
+        type: 'string',
+      },
     },
     getStepStartStates,
     integrationSteps: [
--- a/docs/development.md
+++ b/docs/development.md
@@ -131,6 +131,8 @@ gcloud services enable \
   accesscontextmanager.googleapis.com
 ```
 
+NOTE: This may need broken up into multiple commands. The max batch size is 20.
+
 ### Creating Google Cloud project service account
 
 A
@@ -221,6 +223,7 @@ SERVICE_ACCOUNT_KEY_FILE={...}
 PROJECT_ID="my-j1-proj"
 ORGANIZATION_ID=""
 CONFIGURE_ORGANIZATION_PROJECTS=false
+FOLDER_ID""
 ```
 
 #### `SERVICE_ACCOUNT_KEY_FILE` {string}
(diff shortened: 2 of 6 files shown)


eXtremeX (dismissed):
Nice work! It does what INT-2318 asked for. :rocket: 
I'll mark it as approved and I'll leave a few comments - just some of my thoughts.

A few comments:
(1) Just a reminder - if anyone wants to test this (by running tests), the `test/config.ts` will also need to have mock `integrationConfig` updated with `folderId` value. (_Could be a small gotcha to our future selves, maybe it'd be good if we put some value in there, then again maybe it's unnecessary considering there aren't tests that test this specifically at this point_)

(2) It's very clear from reading the INT-2318 that this should only ingest children folders (not the `folderId` specified as well, plus it's also clear from the code as we're calling the `getAllInnerFolders` and passing it the `folderId`), however it might not be that obvious to the other customers (other than the one asking for this), so if we want to have this as a general change (that other customers could use too) it might be a good idea to document this.

Perhaps in the CHANGELOG.md or ideally as the `folderId`'s description (`integration-google-cloud`'s `integrationDefinition` `configField`'s `description` field perhaps - a mouthful :grimacing:) so that frontend will pick it up and display it as a description next to it in the integration config screen).

src/index.ts
```diff
@@ -51,6 +51,9 @@ export const invocationConfig: IntegrationInvocationConfig<IntegrationConfig> =
         type: 'boolean',
         mask: false,
       },
+      folderId: {
```
austinkelleher, line 54: Should we make this a comma separated string? Do you think that users will want to supply multiple folders?

adam-in-ict: Talking with Tony about this, this customer only wanted to supply a single folder to target.  Plus, the more I looked at it, the more it seems that this integration has been built with the intent being multiple integration instances instead of one large overall integration.  If they have multiple business units in separate distinct folders, they could define multiple "main" integrations that handle creation of project ingestion instances for each folder area.

That said, I'm not working with perfect knowledge and am happy to go back to our previously discussed plan to make this a comma separated list.

austinkelleher: Great feedback. This makes complete sense to me. It would be nice to document the reasoning somewhere!

austinkelleher: Perhaps some of the thoughts you described should show up in the J1 public documentation? What do you think?

adam-in-ict: 100% agree.  I'm planning to push an update to development.md shortly with some comments in it.  @austinkelleher, I may need a refresher on the latest process that gets our documentation out to the J1 public documentation area, since I believe it has changed semi-recently.

austinkelleher (approved):

PR #419, opened 2022-01-31
INT-2448 - Change cloud run service key and its typeAndKeyFromResourceIdentifier

eXtremeX:
This PR fixes the bug associated with Cloud Asset's `createApiServiceToAnyResourceRelationships` step trying to process `iamBinding` that is related to Cloud Run service.

A different key format was used for Cloud Run Service than what the above-mentioned step was searching for, resulting in it failing to find it and ending up creating multiple mapped relationships where it ultimately failed with "Duplicate _key detected" error.

I had to introduce [cacheCloudRunServiceKeyAndUid](https://github.com/JupiterOne/graph-google-cloud/blob/fix/cloud-asset-cloud-run-key-bug/src/utils/jobState.ts#L29) since CloudRunRoutes and CloudRunConfigurations needed to find the CloudRunService, however, they have the `uid` and not all of the necessary parameters for creating the new key. This allows them to exchange the uid for the key.

To save some duplicate code/calls, I made the cacheCloudRunServiceKeyAndUid also return the key so that it can immediately be used [here](https://github.com/JupiterOne/graph-google-cloud/blob/fix/cloud-asset-cloud-run-key-bug/src/steps/cloud-run/index.ts#L51). I like it and dislike it at the same time, perhaps it's better if I just regenerate it again without trying anything fancy? My concern is people won't usually expect this function to return a value (especially with its current name - I could also add "AndReturnsKey" but it's already long as is), that said it saves few lines of code, and even if you do nothing with the returned value it still works as expected.

Other changes include choosing [different typeToKey](https://github.com/JupiterOne/graph-google-cloud/blob/fix/cloud-asset-cloud-run-key-bug/src/utils/iamBindings/typeToKeyGeneratorMap.ts#L117) for Cloud Run Service.

### DIFF
--- a/src/steps/cloud-run/converters.ts
+++ b/src/steps/cloud-run/converters.ts
@@ -20,8 +20,12 @@ export interface MetadataComputedPropertyData {
   duplicateProperties: string[];
 }
 
-export function getCloudRunServiceKey(uid: string) {
-  return `cloudrun_service:${uid}`;
+export function getCloudRunServiceKey(
+  projectId: string,
+  location: string,
+  name: string,
+) {
+  return `projects/${projectId}/locations/${location}/services/${name}`;
 }
 
 export function getCloudRunRouteKey(uid: string) {
@@ -35,6 +39,7 @@ export function getCloudRunConfigurationKey(uid: string) {
 export function createCloudRunServiceEntity(
   data: run_v1.Schema$Service,
   projectId: string,
+  key: string,
 ) {
   // Build webLink
   let webLink = '';
@@ -54,7 +59,7 @@ export function createCloudRunServiceEntity(
       assign: {
         _class: ENTITY_CLASS_CLOUD_RUN_SERVICE,
         _type: ENTITY_TYPE_CLOUD_RUN_SERVICE,
-        _key: getCloudRunServiceKey(data.metadata?.uid as string),
+        _key: key,
         name: data.metadata?.name,
         function: ['workload-management'],
         displayName: data.metadata?.name as string,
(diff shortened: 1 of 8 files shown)


src/steps/cloud-run/converters.ts
```diff
@@ -20,8 +20,12 @@ export interface MetadataComputedPropertyData {
...
 }
 
-export function getCloudRunServiceKey(uid: string) {
-  return `cloudrun_service:${uid}`;
+export function getCloudRunServiceKey(
+  projectId: string,
+  location: string,
+  name: string,
+) {
+  return `projects/${projectId}/locations/${location}/services/${name}`;
```
austinkelleher, line 28: 👍

src/utils/jobState.ts
```diff
@@ -25,6 +26,32 @@ export async function getProjectNameFromId(
...
+): Promise<string> {
+  const location = cloudRunService.metadata?.labels
+    ? cloudRunService.metadata?.labels['cloud.googleapis.com/location']
+    : '';
+  const name = cloudRunService.metadata?.name;
+
+  const uid = cloudRunService.metadata?.uid;
+  const key = getCloudRunServiceKey(projectId, location, name as string);
+
+  await jobState.setData(`cloudRunServiceKey:${key}`, uid);
```
austinkelleher, line 42: Are we fetching this from anywhere? I only see the reference to fetching `cloudRunServiceUid`

eXtremeX: Are you asking if we're ever exchanging `key` for the `uid`? If so, yes we aren't you're right about this!

I was contemplating adding in this capability (since it won't be used, at least for now), still ended up adding it for the sake of "_we might need it at some point and I could write this function more generically while at it_", but it is indeed one additional (useless) `jobState.setData()` call.

Should I get rid of it?

austinkelleher: I see! Let's get rid of it until we need it. I don't want to consume memory on the machine unnecessarily. Even if it's just a small amount. If you think there is a valid use case, please create a Jira ticket that describes the follow-up changes.

eXtremeX: Sounds good, I'll remove it! :+1:

austinkelleher (approved):

PR #421, opened 2022-02-09
INT-2425 - Use organization policies to infer buckets public state

eXtremeX:
This PR adds the ingestion of organization policies which enables us to check for `storage.publicAccessPrevention` constraint.
The policy (and whether it was found / or not / or step failed) will affect every bucket `public` property.

I've discussed this in depth with @austinkelleher already and left a fair bit of comments in the code that should help explain the flow, let me know if something is unclear! Looking forward to hearing your thoughts on this one.


src/steps/orgpolicy/client.ts
```diff
@@ -0,0 +1,32 @@
...
+    parent: string,
+    callback: (
+      data: orgpolicy_v2.Schema$GoogleCloudOrgpolicyV2Policy,
+    ) => Promise<void>,
+  ): Promise<void> {
+    const auth = await this.getAuthenticatedServiceClient();
+
+    await this.iterateApi(
+      async (nextPageToken) => {
+        return this.client.organizations.policies.list({
```
ndowmon, line 17: Since we're looking for a very _specific_ policy here (`storage.publicAccessPrevention`), could we not just call `this.client.organizations.policies.get()`?

src/steps/orgpolicy/index.ts
```diff
@@ -0,0 +1,208 @@
...
+    });
+  }
+
+  return {
+    organizationPolicy,
+    result,
+  };
+}
+
+export async function fetchOrganizationPolicies(
```
ndowmon, line 154: I'd like to see us use a more specific name here, since we're not fetching all of the policies, but rather one very specific policy. What about `fetchOrganizationPublicAccessPreventionPolicy()`?

Or perhaps better yet, what if we simply fetch this policy in a `try/catch` in the bucket step itself? Since as far as I can tell, the bucket step is the only place this gets used.

src/steps/orgpolicy/index.ts
```diff
@@ -0,0 +1,208 @@
...
+): Promise<OrgPolicyResult> {
+  let result = NearestOrgPolicyResult.NOT_FOUND;
+  let organizationPolicy:
+    | orgpolicy_v2.Schema$GoogleCloudOrgpolicyV2Policy
+    | undefined = undefined;
+
+  await client.iterateOrganizationPolicies(
+    `projects/${client.projectId}`,
+    async (orgPolicy) => {
+      // If policy has "inherit" value, it won't be returned here, so we don't need to check if it's "inherit"|"enforced=true"|"enforced=false"
```
ndowmon, line 80: This comment about `inherit` is interesting to me... it actually seems like it'd be _ideal_ if we made just one API call to fetch `storage.publicAccessPrevention` and got either the value for this project _or_ the value **inherit**ed from its nearest parent. How can we make that happen?

src/steps/orgpolicy/index.ts
```diff
@@ -0,0 +1,208 @@
...
+  if (organizationPolicy) {
+    return {
+      organizationPolicy,
+      result,
+    };
+  }
+
+  // A helper function that prepares data
+  // ex: folderPaths = [ { parent: "organizations/958457776463", name: "folders/306329820365", projects: [ 'j1-gc-integration-dev-v3' ] }, ...]
+  const folderPaths = await getFolderPaths(rmClient);
```
ndowmon, line 101: I feel like it'd be easier to just start at the project and work your way up rather than starting at the org and working your way down. Something like this?

```ts
let name = `projects/${this.projectId}`;

while (name) {
  const publicAccessPreventionPolicy = await getPublicAccessPreventionPolicyForResource(name);
  if (publicAccessPreventionPolicy) {
    return {
      organizationPolicy: publicAccessPreventionPolicy,
      result: NearestOrgPolicyResult.FOUND,
    }
  }
  name = await getParent(name);
}
```

Where `getParent` is something like:

```ts
async function getParent(name: string) {
  if (isProject(name)) {
    const project = await this.client.projects.get({ auth, name });
    return project.data.parent;
  } else if (isFolder(name)) {
    const folder = await this.client.folders.get({ auth, name });
    return folder.data.parent;
  } else if (isOrganization(name)) {
    return undefined; // orgs are the root and can have no parent
  } else {
    throw new Error('getParent should receive either a project, folder, or organization.')
  }
}
```

ndowmon (commented):
Hey @eXtremeX, thanks for setting this up! I'd love to hear your thoughts on some of my suggestions.

src/steps/orgpolicy/index.ts
```diff
@@ -0,0 +1,208 @@
...
+    organizationPolicy: <organizationPolicyData>
+
+  a): The enforced policy was 'true' (enforce preventPublicAccess)
+   This overwrites any other public state properties and we know the buckets aren't public.
+  b): The enforced policy was 'false' (enforce preventPublicAccess)
+   This means we have to trust other public state properties (previous code).
+  */
+
+  await jobState.setData(
+    'organization_policy:public_access_prevention',
```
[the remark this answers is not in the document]
eXtremeX: Sure thing, that's a good idea!

src/steps/storage/index.ts
```diff
@@ -56,13 +61,14 @@ export async function fetchStorageBuckets(
...
-      createCloudStorageBucketEntity({
-        data: bucket,
-        projectId: config.serviceAccountKeyConfig.project_id,
-        isPublic: bucketPolicy && isBucketPolicyPublicAccess(bucketPolicy),
-      }),
-    );
+    const bucketEntity = createCloudStorageBucketEntity({
+      data: bucket,
+      projectId: config.serviceAccountKeyConfig.project_id,
+      isPublic: bucketPolicy && isBucketPolicyPublicAccess(bucketPolicy),
```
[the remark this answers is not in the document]
eXtremeX: Agree, the introduction of this 3rd piece makes everything slightly more confusing, I'll make that change :+1:

src/steps/storage/converters.ts
```diff
@@ -17,18 +18,52 @@ export function getCloudStorageBucketKey(id: string) {
   return `bucket:${id}`;
 }
 
+function getPublicState(
```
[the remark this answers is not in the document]
eXtremeX: I didn't actually, I thought AWS repo is private but I was aware of efforts in porting/creating new AWS repo that's more modern, still I'm very much "not up to date" when it comes to that one either.

Alright, so it looks very similar but GCP still has that policy that can shadow both of those values. I can try to make code look a bit more like in AWS, though we'll still need to also apply policy and let it change it. :+1:

src/steps/orgpolicy/client.ts
eXtremeX: Makes sense, it wasn't super clear if we're going to ingest every policy or just this one when I began the work on this piece and that's reason it stayed as `.list`. Totally agree that unless we switch that, using `.get` makes more sense.

src/steps/orgpolicy/index.ts
eXtremeX: Regarding the new name :+1: 

Regarding the other comment - I'm a bit on the fence since what we're currently doing is we start trying to get policy for the project, and until we find it, we keep searching parent folders until we hit the organization, give that one a search too and by that point we have either found it or not. If I move the first part of that logic to the buckets section (searching project for its `storage.publicAccessPrevention` policy) in scenario where we didn't find it, would you (only) then call the outside code (similar to one there's now) to check other folders/organization? That could work but I'm personally finding the way it's currently set up a bit more clearer/easier perhaps since we know there's a dedicated step for this and it has only one job (to find it and store it in jobState). What you think about this?

src/steps/orgpolicy/index.ts
eXtremeX: There may be a slight misunderstanding, let's see.

If you make API call to fetch `storage.publicAccessPrevention` for the project (e.g. "at the project level"), the response value could either be: **true**, **false**, or **[nothing]** - nothing would imply that the policy wasn't set and that instead it (by default) inherits the parent's policy (if any).

We can't stop there because a folder above could have the policy set (with values either set to true, false, or again [nothing] - meaning we need to keep searching higher and higher) and we're interested in finding it since the effects cascades down to our buckets in questions.

That means we can't ever make just one API call (unless we're lucky and it resulted in policy being set at the project level).

Fwiw there's a high chance I misunderstood you and you were aware of what I wrote above, let me know!

src/steps/orgpolicy/index.ts
ndowmon: > If I move the first part of that logic to the buckets section (searching project for its storage.publicAccessPrevention policy) in scenario where we didn't find it, would you (only) then call the outside code (similar to one there's now) to check other folders/organization?

My suggestion is basically to move the entire thing - checking the project, then checking the parent folders, and finally checking the organization. 

> That could work but I'm personally finding the way it's currently set up a bit more clearer/easier perhaps since we know there's a dedicated step for this and it has only one job

I'm good with this as well! Just wanted to get your thoughts on a different option.

src/steps/orgpolicy/index.ts
ndowmon: > If you make API call to fetch `storage.publicAccessPrevention` for the project (e.g. "at the project level"), the response value could either be: **true**, **false**, or **[nothing]** - nothing would imply that the policy wasn't set and that instead it (by default) inherits the parent's policy (if any).

I suppose the very presence of `inherit` as an option here makes me wonder whether it's possible to make a _single_ API call to determine the project's nearest `storage.publicAccessPrevention` configuration. Similar to how listing IAM permissions for a project will return any permissions assigned directly to that project or inherited from a higher-level scope.

I believe you've been hitting the `list` endpoint for policies, which clearly doesn't return any policies with the `inherit` property, but is it possible that the calling the `get` endpoint specifically for `storage.publicAccessPrevention` would actually allow you to get the parent policy with the `inherit` property?

ndowmon: Or maybe the `getEffectivePolicy` would do it? Here's all the possibly related endpoints I've found in the client.

- this.client.projects.policies.get
- this.client.projects.policies.getEffectivePolicy
- this.client.folders.policies.get
- this.client.folders.policies.getEffectivePolicy
- this.client.organizations.policies.get
- this.client.organizations.policies.getEffectivePolicy

src/steps/storage/index.ts
```diff
@@ -37,6 +23,20 @@ export async function fetchStorageBuckets(
...
+
+  try {
+    publicAccessPreventionPolicy =
+      await orgPolicyClient.fetchOrganizationPublicAccessPreventionPolicy();
+  } catch (err) {
+    logger.publishEvent({
+      name: 'missing_permission',
+      description:
+        '"orgpolicy.policies.get" is not a required permission to run the Google Cloud integration, but is required for getting organization policy for "storage.publicAccessPrevention"',
+    });
```
ndowmon, line 38: Does `docs/jupiterone.md` say anything about adding the `orgpolicy.policies.get` permission? If not, can you add it? Also, it sort of seems like this permission is actually `orgpolicy.policy.get`, not `orgpolicy.policies.get`:
Screen Shot 2022-03-01 at 12 05 58 PM

ndowmon (dismissed):

src/steps/orgpolicy/client.ts
```diff
@@ -0,0 +1,20 @@
...
+import { Client } from '../../google-cloud/client';
+
+export class OrgPolicyClient extends Client {
+  private client = google.orgpolicy('v2');
+
+  async fetchOrganizationPublicAccessPreventionPolicy(): Promise<
+    boolean | undefined
+  > {
+    const auth = await this.getAuthenticatedServiceClient();
+    const resp = await this.client.projects.policies.getEffectivePolicy({
```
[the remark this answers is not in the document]
eXtremeX: It is, thanks to your :guitar: suggestions!

src/steps/storage/index.ts
eXtremeX: Was about to fix this, then noticed this: https://cloud.google.com/orgpolicy/docs/reference/rest/v2/projects.policies/getEffectivePolicy

It looks like it's indeed `orgpolicy.policies.get`, but I fully agree that this is/was confusing since there also exists `orgpolicy.policy.get` :man_shrugging: 

My guess is - `orgpolicy.policy.get` is reserved for when you just want to get a single policy via its id, but since getEffectivePolicy needs to "list" policies behind the scenes to find the effective one they decided to mix the `policies` and `get`. It's just a guess though.

I'll update the docs.

ndowmon (approved):

PR #427, opened 2022-03-08
Decouple SQL admin steps from keyvault steps

ndowmon:
### Changed

- Decoupled `fetch-sql-admin-instances` step from key vault dependencies. This
  allows the SQL admin step to execute successfully even if key vault steps are
  disabled.


austinkelleher (dismissed):

mdaum (dismissed):
LGTM

mdaum (approved):

test/recording.ts
```diff
@@ -166,6 +169,36 @@ function getRedactedOAuthResponse() {
...
+
+export function getMatchRequestsBy(
+  instanceConfig: IntegrationConfig,
+): MatchRequestsByFn {
+  return {
+    headers: false,
+    body: false,
+    url: {
+      pathname: (pathname: string): string => {
+        pathname = pathname.replace(
```
ndowmon, line 192: You'll notice that many of our recordings hard-code the `projectId` property so that polly recordings match even if called with a different set of `instanceConfig` credentials.

This `getMatchRequestsBy` automatically normalizes the `projectId` so that we no longer need to hard-code anything in our recording configurations. The pattern is used extensively in the `graph-azure` project.

src/steps/sql-admin/index.test.ts
```diff
@@ -173,13 +176,13 @@ describe('#fetchSQLInstances', () => {
...
 
-describe('#fetchSQLInstances encrypted', () => {
+describe('buildSqlAdminInstanceKmsKeyRelationships', () => {
   let recording: Recording;
 
   beforeEach(() => {
     recording = setupGoogleCloudRecording({
       directory: __dirname,
-      name: 'fetchSQLInstances-encrypted',
+      name: 'buildSqlAdminInstanceKmsKeyRelationships',
```
ndowmon, line 185: I was able to just rename/repurpose the existing `fetchSQLInstances-encrypted` recording, since they hit the same APIs.

src/getStepStartStateExtraTests.test.ts
```diff
@@ -2,9 +2,14 @@ import {
   createMockExecutionContext,
   Recording,
 } from '@jupiterone/integration-sdk-testing';
-import { IntegrationConfig } from '.';
+import { validateStepStartStates } from '@jupiterone/integration-sdk-runtime/dist/src/execution/validation';
```
ndowmon, line 5: It would be nice if `@jupiterone/integration-sdk-testing` had a top-level export to help us validate the `getStepStartStates` function. Not loving the deep export from `@jupiterone/integration-sdk-runtime`.

test/recording.ts
```diff
@@ -166,6 +169,36 @@ function getRedactedOAuthResponse() {
   };
 }
 
+export async function withGoogleCloudRecording(
```
ndowmon, line 172: @austinkelleher you may like the look of this.

PR #428, opened 2022-03-08
INT-2772 - Add GKE labels as properties

eXtremeX:
This PR adds GKE's labels/tags/description to the entities.

Here's a helpful screenshot of console (for a comparison against snapshots):
Screenshot from 2022-03-04 16-14-43

Few gotchas:
I've set `additionalProperties: true` for GKE clusters/nodes tests - there is an arbitrary number of properties now coming our way (these labels) so this acts as a "catch-all". Now, this isn't ideal either, my options seem to be:

1) Simply expect the labels as they're set up in this example cluster (will be painful if this step code ever gets modified by someone and they need to recreate clusters in this exact shape and form).
2) [Best] - modify `toMatchGraphObjectSchema` to allow something like `"metadata.*": { type: 'string' }` if that already doesn't exist? 

Another issue - auto-formatting destroyed the look of changelog I was going for, therefore I'd like to ask if I should keep changelog extremely simple (e.g. 1 liner) or if we should try to explain it a bit better in there (maybe I could use the table-view).

image

Included are also some comments in the converter that better explain how these values are shown in the GCP console (that was the inspiration for choosing how to construct these dot-separated keys).


eXtremeX:
I've noticed the conflict for CHANGELOG.md, but since I need to modify it anyway (the current formatted look is silly), I'll wait for some thoughts on that and then fix the conflicts.

src/steps/containers/index.test.ts
```diff
@@ -118,7 +119,7 @@ describe('#fetchContainerClusters', () => {
       ).toMatchGraphObjectSchema({
         _class: ['Group'],
         schema: {
-          additionalProperties: false,
+          additionalProperties: true,
```
ndowmon, line 122: > I've set `additionalProperties: true` for GKE clusters/nodes tests - there is an arbitrary number of properties now coming our way (these labels) so this acts as a "catch-all".

I think what you've done here with `additionalProperties: true` is perfectly reasonable.

ndowmon (approved):
Looks good to me! Thanks!

src/steps/containers/index.test.ts
eXtremeX: That's a relief, nice to hear that!

src/steps/containers/converters.ts
```diff
@@ -13,6 +13,20 @@ export function getContainerClusterKey(clusterId: string) {
...
+  labels: { [key: string]: string } | undefined | null,
+) {
+  if (!labels) {
+    return;
+  }
+  const prefixed = Object.entries(labels).map(([key, value]) => [
+    `${prefix}.${key}`,
+    value,
+  ]);
+  return prefixed.reduce((acc, [key, value]) => ({ ...acc, [key]: value }), {});
```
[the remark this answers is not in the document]
eXtremeX: That crossed my mind I was just worried about it looking a bit too cryptic, but if it's fine I'm all up for combining that :+1:

ndowmon (approved):

PR #433, opened 2022-03-14
Int 2852 decouple disks kms

eXtremeX:


src/steps/compute/index.ts
```diff
@@ -380,20 +381,28 @@ export async function fetchComputeDisks(
...
   await jobState.iterateEntities(
     { _type: ENTITY_TYPE_COMPUTE_DISK },
-    async (
-      diskEntity: {
-        kmsKeyName?: string | null | undefined;
-      } & Entity,
-    ) => {
-      if (!diskEntity.kmsKeyName) {
+    async (diskEntity) => {
+      const instance = getRawData<compute_v1.Schema$Disk>(diskEntity);
```
ndowmon, line 389:
```suggestion
      const disk = getRawData<compute_v1.Schema$Disk>(diskEntity);
```

ndowmon (approved):
I think the `instance` variable is mis-named, but this isn't blocking.

src/steps/compute/index.ts
eXtremeX: Just want to leave a comment here, we've talked about this over Slack and I simply decided to use `instance` everywhere (instance - an instance of raw data was how I looked at it), so we've decided it's fine to just keep instance everything right now.

PR #435, opened 2022-03-15
Int 2856 decouple dataproc kms

eXtremeX:


src/steps/dataproc/index.ts
```diff
@@ -52,9 +80,26 @@ export async function fetchDataprocClusters(
             to: kmsKeyEntity,
           }),
         );
+      } else {
```
ndowmon, line 83: Can you remove this `else` statement? I'd like to keep the logic as close to the original as possible.

I'm also a little suspicious about whether the key can be from outside of this project, but then again I am a suspicious guy.

ndowmon (approved):

austinkelleher (approved):

PR #442, opened 2022-03-16
Int 2855 decouple logging sinks storage

eXtremeX:


ndowmon (commented):
Left a comment about `destination.includes()`

ndowmon (approved):

PR #445, opened 2022-03-18
Bump @jupiterone/integration-sdk-*@8.6.3

ndowmon:
Because of an update to `.toMatchGraphObjectSchema`, some tests needed to be updated so that the matcher's `_class` argument (which was previously string, like `'Service'`, etc) matched the entity's `_class` value (which is an array, like `['Service']`, etc)


austinkelleher (approved):
Thanks!

PR #454, opened 2022-03-25
Increas serviceusage.services.list page size from 50 -> 200

ndowmon:
### Changed

- Increase `serviceusage.services.list` page size from 50 -> 200


src/steps/service-usage/index.test.ts
```diff
@@ -25,7 +25,7 @@ describe('#fetchApiServices', () => {
 
   test('should collect data', async () => {
```
ndowmon, line 26: The tests setup is pretty stale and I needed to disable 12 tests. I made sure that at least the one test that relies on this API call was successful, but I'm taking time to re-enable the rest of the tests now.

src/steps/service-usage/client.ts
```diff
@@ -23,6 +23,7 @@ export class ServiceUsageClient extends Client {
         return this.client.services.list({
           parent: `projects/${this.projectId}`,
           auth,
+          pageSize: 200,
```
ndowmon, line 26: This is the change.

src/steps/service-usage/index.test.ts
ndowmon: Re-enabled 6 of 12 tests.

austinkelleher: @ndowmon Can you please open an issue to track unskipping these tests?

austinkelleher (approved):

src/steps/service-usage/index.test.ts
ndowmon: Re-enabled 9 of 12 tests

PR #459, opened 2022-03-29
INT-3137 - Calculate correct value of `blockProjectSSHKeys`

austinkelleher:
Properly calcualte the value of `blockProjectSSHKeys` on
`google_compute_instance` entities. According to the GCP
documentation, `block-project-ssh-keys` generally has the
value `TRUE` instead of `true`. The new implementation
considers both cases.

For more information, see: https://cloud.google.com/compute/docs/connect/restrict-ssh-keys#during-vm-creation

Additionally, the logic that was considering compute
instances that have names beginning with `gke-` has been
removed. We should maintain the original value of the
compute metadata on the `google_compute_instance` entity and
let the managed question handle the CIS benchmark logic.


src/steps/compute/converters.ts
```diff
@@ -384,27 +384,17 @@ function getDefaultServiceAccountUsage(
 }
 
 // 4.3 Ensure "Block Project-wide SSH keys" is enabled for VM instances (Scored)
-function getBlockProjectSSHKeysValue(
-  name: string,
+export function getBlockProjectSSHKeysValue(
```
austinkelleher, line 387: The main changes are in this function.

src/steps/compute/converters.ts
```diff
@@ -384,27 +384,17 @@ function getDefaultServiceAccountUsage(
...
-    We usually don't want to map original values, however this one returns true/false as a string
-    It might be confusing if JupiterOne contained fields where true/false can be found in both string and boolean forms?
-   */
   const value = metadata.items?.find(
     (item) => item.key === 'block-project-ssh-keys',
   )?.value;
 
-  return value ? value === 'true' : undefined;
+  // Handle both `TRUE` and `true`
+  return value ? value.toLowerCase() === 'true' : false;
```
austinkelleher, line 397: This is the main fix. `toLowerCase` the node name to handle `TRUE` as well.

src/steps/compute/converters.ts
```diff
@@ -384,27 +384,17 @@ function getDefaultServiceAccountUsage(
...
-function getBlockProjectSSHKeysValue(
-  name: string,
+export function getBlockProjectSSHKeysValue(
   metadata?: compute_v1.Schema$Metadata,
 ): boolean | undefined {
-  /*
-    Exception:
-    VMs created by GKE should be excluded. These VMs have names that start with gke- and are labeled goog-gke-node
-  */
-  if (name.startsWith('gke-') || !metadata) {
```
austinkelleher, line 395: Removed this. See PR description for more info why.

eXtremeX (approved):
Looks great, thanks for the description!

PR #463, opened 2022-04-01
INT-3172 - Add better handling for app engine services

eXtremeX:
The problem is described in details in [Jira task](https://jupiterone.atlassian.net/browse/INT-3172).

### DIFF
--- a/src/steps/app-engine/index.ts
+++ b/src/steps/app-engine/index.ts
@@ -212,8 +212,12 @@ export async function fetchAppEngineServices(
     logger,
   } = context;
 
-  const appEngine = await jobState.getData(ENTITY_TYPE_APP_ENGINE_APPLICATION);
-  if (!appEngine) {
+  // There's 1:1 mapping of GCloud project to AppEngine application
+  // So we can safely map these to the only one existing app engine application
+  const appEngineEntity = await jobState.getData(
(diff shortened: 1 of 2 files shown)


src/steps/app-engine/index.ts
```diff
@@ -212,6 +212,18 @@ export async function fetchAppEngineServices(
     logger,
   } = context;
 
+  // There's 1:1 mapping of GCloud project to AppEngine application
+  // So we can safely map these to the only one existing app engine application
+  const appEngineEntity = await jobState.getData(
+    ENTITY_TYPE_APP_ENGINE_APPLICATION,
+  );
+  if (!appEngineEntity) {
```
austinkelleher, line 220: It looks like this is only used for the relationship construction. 

If we exit early, we will not create any of the App Engine service entities here: https://github.com/JupiterOne/graph-google-cloud/blob/2a9dbf22e132a788604fbdeb8b5be3bec8dc780f/src/steps/app-engine/index.ts#L233

Is that ideal?

austinkelleher (approved):

src/steps/app-engine/index.ts
eXtremeX: Note: This got approved because app engine services can't exist without an app engine application. Thanks to that we (thankfully) won't be missing those Entities even if it seemed that way at first.

PR #480, opened 2022-04-23
INT-3496 - Fix managed question to use correct `_type` query

austinkelleher:
The managed question `good` query is using the correct `_type`,
while the `bad` query is not. Both queries should be using
`google_user` instead of `google_iam_service_account`.

Additionally, both `good` and `bad` queries have been optimized
to use `WITH` query filters instead of `WHERE`, since `WITH`
performs post-traversal filtering.

### DIFF
--- a/jupiterone/questions/questions.yaml
+++ b/jupiterone/questions/questions.yaml
@@ -204,25 +204,27 @@ questions:
   queries:
   - name: good
     query: |
-      find google_user as user
-        that ASSIGNED google_iam_binding
-        that USES google_iam_role as role
-        where
-          role.name!="roles/iam.serviceAccountUser" and
-          role.name!="roles/iam.serviceAccountTokenCreator"
-      return
-        user.name,
+      FIND google_user AS user
+        THAT ASSIGNED google_iam_binding
+        THAT USES google_iam_role
+          WITH (
+            role.name != "roles/iam.serviceAccountUser" AND
+            role.name != "roles/iam.serviceAccountTokenCreator"
+          ) AS role
+      RETURN
+        user.username,
         role.name
   - name: bad
     query: |
-      find google_iam_service_account as user
-        that ASSIGNED google_iam_binding
-        that USES google_iam_role as role
-        where
-          role.name="roles/iam.serviceAccountUser" or
-          role.name="roles/iam.serviceAccountTokenCreator"
-      return
-        user.name,
+      FIND google_user AS user
+        THAT ASSIGNED google_iam_binding
+        THAT USES google_iam_role
+          WITH (
+            role.name ="roles/iam.serviceAccountUser" OR
+            role.name ="roles/iam.serviceAccountTokenCreator"
+          ) AS role
+      RETURN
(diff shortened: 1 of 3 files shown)


jupiterone/questions/questions.yaml
```diff
@@ -204,25 +204,27 @@ questions:
...
     query: |
-      find google_iam_service_account as user
-        that ASSIGNED google_iam_binding
-        that USES google_iam_role as role
-        where
-          role.name="roles/iam.serviceAccountUser" or
-          role.name="roles/iam.serviceAccountTokenCreator"
-      return
-        user.name,
+      FIND google_user AS user
```
austinkelleher, line 219: Updated `_type` from `google_iam_service_account` -> `google_user`

jupiterone/questions/questions.yaml
```diff
@@ -204,25 +204,27 @@ questions:
...
-        that USES google_iam_role as role
-        where
-          role.name!="roles/iam.serviceAccountUser" and
-          role.name!="roles/iam.serviceAccountTokenCreator"
-      return
-        user.name,
+      FIND google_user AS user
+        THAT ASSIGNED google_iam_binding
+        THAT USES google_iam_role
+          WITH (
```
austinkelleher, line 210: `WHERE` -> `WITH`

jupiterone/questions/questions.yaml
```diff
@@ -204,25 +204,27 @@ questions:
...
-        that USES google_iam_role as role
-        where
-          role.name="roles/iam.serviceAccountUser" or
-          role.name="roles/iam.serviceAccountTokenCreator"
-      return
-        user.name,
+      FIND google_user AS user
+        THAT ASSIGNED google_iam_binding
+        THAT USES google_iam_role
+          WITH (
```
austinkelleher, line 222: `WHERE` -> `WITH`

jupiterone/questions/questions.yaml
```diff
@@ -204,25 +204,27 @@ questions:
...
+            role.name != "roles/iam.serviceAccountUser" AND
+            role.name != "roles/iam.serviceAccountTokenCreator"
+          ) AS role
+      RETURN
+        user.username,
```
austinkelleher, line 215: Updated to return `username` instead of `name` because is the person's actual name instead of a unique identifier.

jupiterone/questions/questions.yaml
```diff
@@ -204,25 +204,27 @@ questions:
...
-        user.name,
+      FIND google_user AS user
+        THAT ASSIGNED google_iam_binding
+        THAT USES google_iam_role
+          WITH (
+            role.name ="roles/iam.serviceAccountUser" OR
+            role.name ="roles/iam.serviceAccountTokenCreator"
+          ) AS role
+      RETURN
+        user.username,
```
austinkelleher, line 227: Updated to return `username` instead of `name` because is the person's actual name instead of a unique identifier.

ceelias (approved):

PR #493, opened 2022-05-30
INT-2289 - Update docs regarding BigQuery permissions

eXtremeX:
For BigQuery:
`bigquery.datasets.get`

This allows integration to see datasets (datasets house models/tables, so this is the starting point). If this permission isn’t set, the integration won’t show an error but the resulting datasets response will just be an empty list (it’s due to how GCP decided to make that particular endpoint work).

However with just that one permission added, the integration won’t be able to process the tables and models fully (it’ll be able to see them but not get all the metadata), hence the following two permissions are recommended:
`bigquery.models.getMetadata`
`bigquery.tables.get`

So, in total the following permissions are required to make full use of our integration:
`bigquery.datasets.get`
`bigquery.models.getMetadata`
`bigquery.tables.get`

### DIFF
--- a/docs/jupiterone.md
+++ b/docs/jupiterone.md
@@ -153,15 +153,14 @@ compute.projects.get
 orgpolicy.policy.get
 ```
 
-For BigQuery, the following permissions are recommended:
+For BigQuery, the following _additional_ permissions are needed to ingest
+BigQuery datasets, models, and tables respectively:
 
 ```
 bigquery.datasets.get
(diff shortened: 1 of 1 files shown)


ndowmon (approved):

eXtremeX:
> It's not excellent that GCP will silently fail in this case. Do you know of any way to actually warn the user if bigquery.datasets.get is not allowed?

We talked about this, and have decided to tackle this as a separate task.

PR #494, opened 2022-05-30
Update BigQuery docs part

eXtremeX:
Same as #493 but for the new docs branch.


austinkelleher (approved):
Approved with suggestion

PR #495, opened 2022-06-12
INT-4087 - Remove service API calculations from `getStepStartStates`

austinkelleher:
Rely on the individual steps to call service APIs instead of
pre-calculating which service APIs are enabled in `getStepStartStates`.
This should help drastically reduce the number of API calls to
`serviceusage.googleapis.com`.


mdaum (dismissed):

mdaum (approved):

src/getStepStartStates.ts
```diff
@@ -194,13 +179,21 @@ export function getOrganizationSteps() {
...
+}
+
 export default async function getStepStartStates(
   context: IntegrationExecutionContext<SerializedIntegrationConfig>,
 ): Promise<StepStartStates> {
   const { instance, logger } = context;
   const { config: serializedIntegrationConfig } = instance;
 
-  validateInvocationConfig(context);
+  validateInvocationConfig(serializedIntegrationConfig);
```
ndowmon, line 196: suggestion for later: it would be nice to use `loadExecutionContext` in this integration so that the config is already deserialized in `validateInvocation` and `getStepStartStates`.

src/getStepStartStates.ts
```diff
@@ -231,354 +224,157 @@ export default async function getStepStartStates(
...
+    },
+    [STEP_BUILD_ADDITIONAL_PROJECT_BUDGET]: { disabled: !masterOrgInstance },
   };
 
   logger.info(
     { stepStartStates: JSON.stringify(stepStartStates) },
     'Step start states',
   );
-  return stepStartStates;
+  return Promise.resolve(stepStartStates);
```
ndowmon, line 379: Why do you need `Promise.resolve` here?

ndowmon (commented):
Looks good to me. Thanks for using a beta release here.

src/getStepStartStates.ts
austinkelleher: Async method that doesn't have any `await` causes eslint to complain.

PR #505, opened 2022-07-28
Add .nvmrc to repository

i5o:
With node version 14


.nvmrc
```diff
@@ -0,0 +1 @@
+v14.20.0
```
JakeFerrero, line 1: I'd potentially just do 14 rather than a specific version of 14.

i5o: Sounds good to me

JakeFerrero (approved):

PR #507, opened 2022-08-15
INT-4920 google_cloud_function data model

ndowmon:
Started working on this spec as part of INT-4584, in which a prospect told us that we didn't ingest data from the **Google Cloud Functions** service. In fact, we did (and do), but I continued the exercise to determine additional relationships that we could add in the future. 

Unfortunately, the 4 additional relationships that we can produce are only available in the v2 api, while the current code is on the v1 api, so we'll need to migrate to the v2 api before enabling this additional resource ingestion.

v1 API: https://cloud.google.com/functions/docs/reference/rest/v1/projects.locations.functions/list
v2 API: https://cloud.google.com/functions/docs/reference/rest/v2/projects.locations.functions/list

Screen Shot 2022-08-16 at 6 14 44 PM


docs/spec/src/steps/services/functions/index.ts
```diff
@@ -0,0 +1,127 @@
...
+    relationships: [
+      {
+        _type: 'google_cloud_function_uses_storage_bucket',
+        _class: RelationshipClass.USES,
+        sourceType: 'google_cloud_function',
+        targetType: 'google_storage_bucket',
+      },
+    ],
+    dependsOn: ['fetch-cloud-functions', 'fetch-cloud-storage-buckets'],
+    implemented: false,
```
andy-foglia, line 65: Just for my own education, does the `implemented` field here also control whether the new documentation / new relationships are visible to users? e.g. When `implemented` is `false`, is there any reference to this in the docs rendered to end users? Or do other things control that visibility?

ndowmon: The `implemented` field only controls whether this spec is included in the `.toImplementSpec()` test. There is no reference to this in docs. Instead, the docs are controlled by what's _actually_ implemented in `./src/*`. 

See https://github.com/JupiterOne/sdk/blob/main/packages/integration-sdk-testing/src/jest.ts#L133-L149 for more details: 

```ts
      /**
       * Used to verify that an implemented `IntegrationInvocationConfig`
       * matches a specified `IntegrationSpecConfig`
       *
       * @example
       * ```ts
       * import { invocationConfig as implementedConfig } from '.';
       * import { invocationConfig as specConfig } from '../docs/spec/src';
       *
       * test('implemented integration should match spec', () => {
       *   expect(implementedConfig).toImplementSpec(specConfig);
       * });
       * ```
       */
      toImplementSpec<T extends IntegrationInstanceConfig>(
        spec: IntegrationSpecConfig<T>,
      ): R;
```

i5o:
Is it worth to upgrade to v2  -for the extra relationships- now or would it be easier for us to work on v1 and then think about v2?

docs/spec/src/steps/services/build/index.ts
```diff
@@ -0,0 +1,204 @@
...
+        _type: 'google_cloud_build',
+        _class: ['Workflow'],
+      },
+    ],
+    relationships: [],
+    implemented: true,
+  },
+  {
+    /**
+     * PROPERTY: google_cloud_build.source.storageSource
```
i5o, line 29: .source.storageSource.bucket

docs/spec/src/steps/services/build/index.ts
```diff
@@ -0,0 +1,204 @@
...
+        sourceType: 'google_cloud_build',
+        targetType: 'google_storage_bucket',
+      },
+    ],
+    dependsOn: ['fetch-cloud-builds', 'fetch-cloud-storage-buckets'],
+    implemented: true,
+  },
+  {
+    /**
+     * PROPERTY: google_cloud_build.source.repoSource
```
i5o, line 48: .repoSource.repoName

docs/spec/src/steps/services/functions/index.ts
```diff
@@ -0,0 +1,127 @@
...
+     * PATTERN: Fetch Entities
+     * REQUIRED PERMISSIONS: cloudfunctions.functions.list
+     */
+    id: 'fetch-cloud-functions',
+    name: 'Cloud Functions',
+    entities: [
+      {
+        resourceName: 'Cloud Function',
+        _type: 'google_cloud_function',
+        _class: 'Function',
```
i5o, line 20: I assume this will be `['Function']`; it looks like this is also already implemented on `steps/functions/`

docs/jupiterone.md
```diff
@@ -166,7 +166,8 @@ See the
 [Google Cloud custom role documentation](https://cloud.google.com/iam/docs/creating-custom-roles#creating_a_custom_role)
 for additional information on how custom roles can be configured and assigned.
 
-NOTE: You may also create a service account using the [`gcloud` CLI](https://cloud.google.com/sdk/gcloud). There is documentation on 
+NOTE: You may also create a service account using the
```
i5o, line 169: Can we please just use the file on `main` to avoid conflicts?

i5o (approved):
Beside my comment on properties names, it looks good to me 🔥

docs/jupiterone.md
ndowmon: I think this was done by my pre-commit hook, because prettier enforces an 80 character limit. Someone else recently committed these changes who does not have prettier configured, so the automation didn't run previously.

docs/spec/src/steps/services/build/index.ts
ndowmon: Thank you! Missed this but added in #549

docs/spec/src/steps/services/build/index.ts
ndowmon: Thank you! Missed this but added in #549

PR #511, opened 2022-08-17
INT-5233 step: fetch-cloud-builds

i5o:
- [x] Implement step
- [x] Testing
- [x] update md documentation


test/mocks.ts
```diff
@@ -2636,3 +2637,7 @@ export function getMockSecretVersionEntity(): secretmanager_v1.Schema$SecretVers
     state: 'ENABLED',
   };
 }
+
+export function getMockCloudBuild(): cloudbuild_v1.Schema$Build {
+  return {};
+}
```
ndowmon, line 2643: What do you think about defining `getMockCloudBuild()` inside of the `src/steps/cloud-build/converters.ts` file, or wherever it's used? I realize that this file currently defines mocks across the entire project, but I'm not exactly sure why that is or whether it's a pattern that we should rethink. Would love to hear your thoughts!

ndowmon:
@i5o - I noticed your description has a checkbox for each entity and relationship in the **Google Cloud Build** service. Instead of putting the entire **Google Cloud Build** service in this PR, what do you think about doing each _step_ in a small, independent PR? In my experience, that has been a great way to move quickly in these projects. IMO, this PR looks **Ready for review**.

src/steps/cloud-build/converters.ts
```diff
@@ -0,0 +1,23 @@
...
+  };
+
+  return createGoogleCloudIntegrationEntity(source, {
+    entityData: {
+      source: source,
+      assign: {
+        _class: CloudBuildEntities.BUILD._class,
+        _type: CloudBuildEntities.BUILD._type,
+        _key: data.name as string,
+        name: data.name,
```
ndowmon, line 19: I notice that the `cloudbuild_v1.Schema$Build` interface has a ton of properties on it, such as `artifacts`, `availableSecrets`, `buildTriggerId`, etc. It's best practice to explicitly add all properties to the `assign` field, where possible. 

You probably already know this, but `assign` accepts only primitive or array of primitive properties (e.g. `string | number | boolean | string[] | number[] | boolean []`). I tend to `assign` in the following way:

```ts
interface Schema$Build {
  // ...
  buildTriggerId?: string | null;
  createTime?: string | null;
  finishTime?: string | null;
  id?: string | null;
  images?: string[] | null;
  // ...
  options?: {
    diskSizeGb?: string | null;
    dynamicSubstitutions?: boolean | null;
    // ...
  }
  // ...
  secrets?: {
    kmsKeyName?: string | null;
    secretEnv?: { [key: string] : string; } | null;
  }[];
  // ...
}

export function createBuildEntity(source: Schema$Build) {
  const data = {
    ...source,
    labels: source.tags,
    tags: undefined,
  };

  return createGoogleCloudIntegrationEntity(source, {
    entityData: {
      source: data,
      assign: {
        _class: CloudBuildEntities.BUILD._class,
        _type: CloudBuildEntities.BUILD._type,
        _key: data.name as string,
        name: data.name,

        buildTriggerId: data.buildTriggerId,
        createTime: parseTimePropertyValue(data.createTime),
        createdAt: parseTimePropertyValue(data.createTime),
        finishTime: parseTimePropertyValue(data.finishTime),
        endedAt: parseTimePropertyValue(data.finishTime),
        id: data.id!,
        images: data.images,
        'options.diskSizeGb': data.options?.diskSizeGb,
        'options.dynamicSubstitutions': data.options?.dynamicSubstitutions,
        // `secrets` is an array of objects... we cannot assign these properties directly.
        // consider something like `'secrets.kmsKeyNames': data.secrets?.map((s) => s.kmsKeyName)
        // or possibly creating a separate entity, depending on the data contained.
      },
    },
  });
}
```

Notes: 
1. use `parseTimePropertyValue()` when creating dates, so that our system interprets different date formats correctly
2. use `createdAt` and `endedAt` tense as well, which is a suggested standard
3. chain object properties with `<object-name>.<property-name>`
4. handle arrays of objects depending on which data exists in the object.

Can you assign the other `Schema$Build` properties to this entity?

ndowmon (commented):
This one looks good and ready-for-review to me. I posted one small comment about the converter, but everything else looks awesome 🔥🔥🔥

ndowmon (approved):
🔥

src/steps/cloud-build/constants.ts
```diff
@@ -0,0 +1,16 @@
...
+    id: 'fetch-cloud-builds',
+    name: 'Fetch Cloud Builds',
+  },
+};
+
+export const CloudBuildEntities = {
+  BUILD: {
+    resourceName: 'Cloud Build',
+    _type: 'google_cloud_build',
+    _class: ['Secret'],
```
i5o, line 12: Whops :(

PR #512, opened 2022-08-19
Expose useEnablementsForStepStartStates config variable

ndowmon:
In the future, we would like _all_ accounts to not require `enablement.getEnabledServiceNames(config)`, because accounts with many GCP integration instances (>1000) quickly hit rate limits at this API.

However, we need to allow different integrations to run with/without `enablement.getEnabledServiceNames(config)` as we confirm that each API handles the correct behavior whether the service is disabled for the the _service account project_ or the _target project_. More details to come.


src/getStepStartStates.ts
```diff
@@ -229,10 +232,46 @@ export default async function getStepStartStates(
...
+      masterOrgInstance,
+      singleProjectInstance,
+      organizationSteps,
+      config,
+      logger,
+    });
+  }
+}
+
+function getDefaultStepStartStates(params: {
```
i5o, line 260: existing one

i5o (approved):
How could I enable `useEnablementsForStepStartStates`, which `.env` variable do I need to set-up?

Otherwise it looks good 👍

austinkelleher (approved):

PR #522, opened 2022-08-29
Update role names

janettelynch:


i5o (approved):
This looks good to me!

I was thinking that we should add the "technical" version of it but it's already on the documentation when you click it.

Thanks for the contribution 😄

PR #524, opened 2022-08-30
INT-5355 upgrade googleapis to support GithubEnterpriseConfigs

i5o:
Support for GithubEnterpriseConfigs was added in v84;

https://github.com/googleapis/google-api-nodejs-client/blob/googleapis-v84.0.0/src/apis/cloudbuild/v1.ts#L870


austinkelleher (approved):
Thanks!

PR #531, opened 2022-09-06
INT-5237 step: fetch-cloud-build-bb-configs

i5o:
Add support for https://cloud.google.com/build/docs/api/reference/rest/v1/projects.locations.bitbucketServerConfigs/list

### DIFF
--- a/src/steps/cloud-build/steps/fetch-cloud-build-bb-configs.ts
+++ b/src/steps/cloud-build/steps/fetch-cloud-build-bb-configs.ts
@@ -1,0 +1,27 @@
+import { IntegrationStep } from '@jupiterone/integration-sdk-core';
+import { IntegrationConfig, IntegrationStepContext } from '../../../types';
+import { CloudBuildClient } from '../client';
+import { CloudBuildEntitiesSpec, CloudBuildStepsSpec } from '../constants';
+import { createGoogleCloudBuildBitbucketServerConfigEntity } from '../converters';
+
+export const fetchCloudBuildBitbucketServerConfigStep: IntegrationStep<IntegrationConfig> =
+  {
+    ...CloudBuildStepsSpec.FETCH_BUILD_BITBUCKET_SERVER_CONFIG,
+    entities: [CloudBuildEntitiesSpec.BUILD_BITBUCKET_SERVER_CONFIG],
+    relationships: [],
+    executionHandler: async function (
+      context: IntegrationStepContext,
+    ): Promise<void> {
+      const {
+        jobState,
+        instance: { config },
+      } = context;
+      const client = new CloudBuildClient({ config });
+
+      await client.iterateBuildsBitbucketServerConfigs(async (data) => {
+        await jobState.addEntity(
+          createGoogleCloudBuildBitbucketServerConfigEntity(data),
+        );
+      });
+    },
+  };
--- a/test/recording.ts
+++ b/test/recording.ts
@@ -3,7 +3,12 @@ import {
   setupRecording,
   SetupRecordingInput,
 } from '@jupiterone/integration-sdk-testing';
-import { cloudfunctions_v1, iam_v1, privateca_v1beta1 } from 'googleapis';
+import {
+  cloudbuild_v1,
+  cloudfunctions_v1,
+  iam_v1,
+  privateca_v1beta1,
+} from 'googleapis';
 import * as querystring from 'querystring';
 import * as url from 'url';
 import { gunzipSync } from 'zlib';
@@ -85,6 +90,12 @@ function isListCertificatesUrl(url: string) {
   ).test(url);
 }
 
+function isListBitbucketServerConfigUrl(url: string) {
+  return new RegExp(
+    /https:\/\/cloudbuild.googleapis.com\/v1\/projects\/(.*?)\/locations\/(.*?)\/bitbucketServerConfigs/,
+  ).test(url);
+}
+
 /**
  * The response from creating a service account key contains the private key
  * data, which we need to redact.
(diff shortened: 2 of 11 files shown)


test/recording.ts
```diff
@@ -85,6 +90,12 @@ function isListCertificatesUrl(url: string) {
   ).test(url);
 }
 
+function isListBitbucketServerConfigUrl(url: string) {
+  return new RegExp(
+    /https:\/\/cloudbuild.googleapis.com\/v1\/projects\/(.*?)\/locations\/(.*?)\/bitbucketServerConfigs/,
```
austinkelleher, line 95: 💯

src/steps/cloud-build/steps/fetch-cloud-build-bb-configs.ts
```diff
@@ -0,0 +1,27 @@
+import { IntegrationStep } from '@jupiterone/integration-sdk-core';
```
austinkelleher, line 1: I'm wondering whether we should rename the `steps` sub-folder to `executionHandlers`, which would be more consistent with the AWS integration. This is what AWS does:

```
src/steps/services/ec2/executionHandlers/networkInterfaces.ts
```

Additionally, maybe we should use camelcase file names to be consistent with AWS as well?

Thoughts @ndowmon @i5o?

austinkelleher (commented):
Looks good! Left one question about folder + file naming consistency since this looks like a new pattern in `graph-google-cloud`.

src/steps/cloud-build/steps/fetch-cloud-build-bb-configs.ts
i5o: Thank you Austin, this is totally up to me.

@ndowmon left his thoughts on the previous MRs:
https://github.com/JupiterOne/graph-google-cloud/pull/518#discussion_r955466032

IMO both work, but let's we should make like AWS does so that we make it an standard

i5o: Just to point out - I think the name `steps` is proper, because we end up exporting the `step`.

I reviewed graph-aws and the difference is that both the `executionHandler` and `step` gets defined and exported in the same file; here we only export the `step` with an inline `executionHandler`.

I know for know it might be enough but in the future we might have huge executionHandlers but anyway I think that sticking with `steps` as the folder is okay -- as it's how we call / refer to them everywhere.

Revisiting proposed filename convention, I'm looking at what already exists, in `graph-google-cloud` we split services names by `-`, for example: `secret-manager`; where in `aws` it would be `secretmanager`, so perhaps it might be better to leave at it is on this pr and the whole project

test/recording.ts
i5o: Thank you.
However I'm not happy about the way this is being done, is this `recording.ts` only for testing?
Maybe we should just redact anything that says "private", "apikey", "webhook", etc?

or even ... find a way to get rid of recordings?

test/recording.ts
```diff
@@ -149,6 +160,21 @@ function sanitizeListCertificatesResponse(
...
+function sanitizeBitbucketServerConfigResponse(
+  response: cloudbuild_v1.Schema$ListBitbucketServerConfigsResponse,
+) {
+  return {
+    ...response,
+    bitbucketServerConfigs: response.bitbucketServerConfigs?.map((config) => {
+      return {
+        ...config,
+        apiKey: '[REDACTED]',
+        webhookKey: '[REDACTED]',
```
ndowmon, line 172: Are these sensitive? Should we perhaps not add them to entities in J1?

test/recording.ts
ndowmon: We can discuss this more IRL if you'd like.

src/steps/cloud-build/steps/fetch-cloud-build-bb-configs.ts
```diff
@@ -0,0 +1,27 @@
...
+        jobState,
+        instance: { config },
+      } = context;
+      const client = new CloudBuildClient({ config });
+
+      await client.iterateBuildsBitbucketServerConfigs(async (data) => {
+        await jobState.addEntity(
+          createGoogleCloudBuildBitbucketServerConfigEntity(data),
+        );
+      });
```
ndowmon, line 25: I have to say... I love how succinct these execution handlers are. 🔥

src/steps/cloud-build/steps/fetch-cloud-build-bb-configs.ts
ndowmon: It's worth having a bit of a deeper discussion on how the directory ought to be structured. Even though I originally advocated for the `/executionHandlers/` directory in `graph-aws`, I don't necessarily love the name. 

I'm perfectly happy with these names for now, though, and don't want to slow up development for this reason.

ndowmon (approved):

test/recording.ts
i5o: By adding you mean excluding them from when we `assign` or just delete the value itself even in `rawdata`?

ndowmon (approved):

src/steps/cloud-build/steps/fetch-cloud-build-bb-configs.ts
```diff
@@ -0,0 +1,27 @@
...
+import { IntegrationConfig, IntegrationStepContext } from '../../../types';
+import { CloudBuildClient } from '../client';
+import { CloudBuildEntitiesSpec, CloudBuildStepsSpec } from '../constants';
+import { createGoogleCloudBuildBitbucketServerConfigEntity } from '../converters';
+
+export const fetchCloudBuildBitbucketServerConfigStep: IntegrationStep<IntegrationConfig> =
+  {
+    ...CloudBuildStepsSpec.FETCH_BUILD_BITBUCKET_SERVER_CONFIG,
+    entities: [CloudBuildEntitiesSpec.BUILD_BITBUCKET_SERVER_CONFIG],
+    relationships: [],
```
ndowmon, line 11: I don't think there's a strong reason to move `entities`, `relationships`, or even `name` to constants, since `id` is the only property that can be shared across different steps / services.

PR #536, opened 2022-09-08
INT-5293 step: build-cloud-build-trigger-triggers-build-relationships

i5o:
quick note on this: `google_cloud_build_trigger.build` in the doc actually refers to a template for the trigger, I’ve instead mapped build with the trigger of said build using `buildTriggerId` field on build.

I still feel like `triggers` is not the right term for this relationship as the build gets `created`

### DIFF
--- a/src/steps/cloud-build/constants.ts
+++ b/src/steps/cloud-build/constants.ts
@@ -40,6 +40,12 @@ export const CloudBuildRelationshipsSpec = {
     _class: RelationshipClass.HAS,
     targetType: CloudBuildEntitiesSpec.BUILD_BITBUCKET_REPO._type,
   },
+  BUILD_TRIGGER_TRIGGERS_BUILD: {
+    _type: 'google_cloud_build_trigger_triggers_build',
+    sourceType: CloudBuildEntitiesSpec.BUILD_TRIGGER._type,
+    _class: RelationshipClass.TRIGGERS,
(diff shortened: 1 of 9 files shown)


src/steps/cloud-build/constants.ts
```diff
@@ -40,6 +40,12 @@ export const CloudBuildRelationshipsSpec = {
     _class: RelationshipClass.HAS,
     targetType: CloudBuildEntitiesSpec.BUILD_BITBUCKET_REPO._type,
   },
+  BUILD_TRIGGER_TRIGGERS_BUILD: {
+    _type: 'google_cloud_build_trigger_triggers_build',
+    sourceType: CloudBuildEntitiesSpec.BUILD_TRIGGER._type,
+    _class: RelationshipClass.TRIGGERS,
```
ndowmon, line 46: Can you propose a different `RelationshipClass` to use here instead?

ndowmon (approved):

src/steps/cloud-build/constants.ts
i5o: I would love to have `RelationshipClass.TRIGGERED`; but `RelationshipClass.CREATED` would work as well.

i5o: We also have `RelationshipClass.GENERATED`

PR #537, opened 2022-09-08
INT-5288 step: build-cloud-build-storage-bucket-relationships

i5o:
Note: This was pushed with --no-verify as I had issues running the fetch cloud storage.
- For testing purposes I had to add `if (bucket.name?.includes('us-central1')) return;` to `storage/index.ts`; once that was added I was able to get the following picture.

image

### DIFF
--- a/test/recording.ts
+++ b/test/recording.ts
@@ -214,10 +214,17 @@ export function getMatchRequestsBy(
     headers: false,
     body: false,
     url: {
+      query: (query) => {
+        if (query['project']) {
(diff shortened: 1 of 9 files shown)


test/recording.ts
```diff
@@ -214,10 +214,17 @@ export function getMatchRequestsBy(
     headers: false,
     body: false,
     url: {
+      query: (query) => {
```
ndowmon, line 217: Yay, thanks!

ndowmon (approved):

PR #545, opened 2022-09-20
Get rid of organizationSetup

i5o:
This is preventing us from removing iterateApi and withErrorHandling from client.test.ts as organizationSetup relies on this exported functions


scripts/organizationSetup.test.ts
```diff
@@ -1,507 +0,0 @@
...
-}
-
-describe('#setupOrganization', () => {
-  test('should iterate only specific projects when projectIds supplied', async () => {
-    await withRecording('setupOrganization', __dirname, async () => {
-      const result = await setupOrganization(
-        getSetupOrganizationParams({
-          projectIds: ['j1-gc-integration-dev'],
-        }),
-      );
```
ndowmon, line 45: Can you also remove the files in `scripts/__recordings__/setupOrganization`

ndowmon (approved):

PR #548, opened 2022-09-22
Int 5556 dev tools 8.24.1

edreyer1:


src/getStepStartStateExtraTests.test.ts
```diff
@@ -2,7 +2,6 @@ import {
   createMockExecutionContext,
   Recording,
 } from '@jupiterone/integration-sdk-testing';
-import { validateStepStartStates } from '@jupiterone/integration-sdk-runtime/dist/src/execution/validation';
```
ndowmon, line 5: Can you provide context on why this code was copied over to this project?

edreyer1: For some reason I was getting a typing error for the parameter of the this function in the last test. I could figure out how to get around it other than adding it here

src/steps/access-context-manager/index.test.ts
```diff
@@ -568,6 +568,7 @@ declare global {
   }
 }
 
+/* eslint-disable no-console */
```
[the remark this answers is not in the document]
edreyer1: Hey I can do this I went ahead and created a ticket for it https://jupiterone.atlassian.net/browse/INT-5632

src/getStepStartStateExtraTests.test.ts
ndowmon: Yeah, can you please do some troubleshooting on the type error? Duplicating this code means that we'll be out-of-sync if any useful changes are introduced in the `@jupiterone/integration-sdk-runtime` package.

ndowmon (approved):

PR #550, opened 2022-09-26
Int 5632 replace toCreateValidRelationshipsToEntities

edreyer1:


src/steps/cloud-asset/index.test.ts
```diff
@@ -45,6 +45,7 @@ import { CLOUD_FUNCTION_ENTITY_TYPE, fetchCloudFunctions } from '../functions';
 import { IAM_MANAGED_ROLES_DATA_JOB_STATE_KEY } from '../../utils/iam';
 import { fetchApiServices } from '../service-usage';
 
+/* eslint-disable no-console */
```
ndowmon, line 48: This file also has a duplicate `toTargetEntities()` matcher. I think we should remove that one from this codebase as well if possible.

ndowmon (approved):

ndowmon (approved):

PR #551, opened 2022-09-30
Account for non-deterministic basic role permissions

ndowmon:
## Background
I was able to determine that the `owner`, `editor`, and `viewer` Google Cloud Roles often churn from one integration invocation to the next. Upon further investigation, I learned that a few specific permissions are _sometimes_ returned from API, and are _sometimes_ not returned.

For example - this evening, I saw arbitrary churn of the following permissions:

- `managedidentities.domains.checkMigrationPermission`
- `managedidentities.domains.disableMigration`
- `managedidentities.domains.enableMigration`

My hunch, given what I’ve seen - GCP is constantly adding new services, endpoints, and permissions to the platform, and thus frequently update permissions on the basic roles (`owner`, `editor` , and `viewer` - >5000, >3000, and >2000 permissions, respectively).

Based on this, I assume that their permissioning system probably includes a service that provides permissions for a given basic role, whose underlying implementation is distributed (because of significant load), and is designed to be **eventually consistent**. Sometimes, making the API call hits a server that has received the most recent permission updates, and sometimes not. 

## Proposed Change
It seems if we make the API call a few times, we would have a higher likelihood of finding the larger (and correct) set of permissions. This PR iterates the managed roles an additional 5 times, remembering which API call returned the largest set of permissions for each of `role/owner`, `role/editor`, and `role/viewer`. It assigns the largest set of permissions to the resulting entity.

## Testing
I have tested this change heuristically - before the change, I saw that, between two subsequent invocations of the `fetch-iam-managed-roles` integration step, at least one of the `owner`, `editor`, or `viewer` roles had a different set of permissions about 75% of the time. _After_ the change, I invoked the same step 10 times in sequence, and saw exactly the same number of permissions on each of the `owner`, `editor`, and `viewer` roles in each of the 10 invocations.

Specifically, I used the following CLI command:
```
yarn start -s fetch-iam-managed-roles -s create-basic-roles --no-cache
```

## Impact
Local testing showed that this change increased the runtime of the `fetch-iam-managed-roles` from ~4 seconds to ~19 seconds. The `roles.list` API returns just 4 pages of data, and since this API lists _predefined roles_, it ought to return just 4 pages of data for _all_ accounts. Hence, it should have a constant impact on all integrations of ~15 seconds, and 20 additional API calls (duplicating 4 pages of data 5 times).

Screen Shot 2022-09-29 at 11 37 33 PM

By hitting the API multiple times and storing the largest set of permissions, we can make the `google_iam_role` and `google_iam_binding` entities more deterministic. This will have a positive impact on the performance of our data pipeline.


src/steps/iam/index.ts
```diff
@@ -168,13 +168,50 @@ export async function fetchIamManagedRoles(
...
-    return Promise.resolve();
   });
 
+  /**
+   * We have learned that GCP basic roles have non-deterministic behavior on
+   * their listed permissions. We hit the same endpoint multiple times to see if
+   * the response includes _additional_ permissions, always choosing the larger set.
+   */
+  for (let i = 0; i < (context.instance.config.isDev ? 0 : 5); i++) {
+    await client.iterateManagedRoles((role) => {
```
philidem, line 192: Maybe just always loop the permissions and add to `Set`? I think that would ensure that you are getting the union returned by all calls.

austinkelleher (dismissed):

kenanwarren (approved):

philidem (approved):

austinkelleher (approved):

PR #558, opened 2022-11-15
INT-5712 Add job logs for steps that will be skipped due to disabled services

JakeFerrero:
We will now print out a job log for all of the disabled google-cloud services
in the customer's account, and the graph steps that will be skipped as a result.
To do this, the interface for createStepStartState has been updated to include
the stepId. A map of disabled services to steps has been added to track which
services are disabled and the steps that will subsequently be disabled as well
as we cerate each step start state.


JakeFerrero:
Ok this is going to look like a lot of changes, but really just the `createStepStartState` function params were updated. However, _every_ step calls this function, so there needed to be a change per step. But the change is just updating what we're passing to the `createStepStartState`. I'd definitely appreciate if someone could take on the arduous task of going through each call and ensuring the step ID's match. I will say though, even if we somehow miss one, the only thing that will be incorrect is the log statement.

JakeFerrero:
This is what the job logs look like:
Screen Shot 2022-11-16 at 11 13 19

src/steps/enablement.ts
```diff
@@ -3,6 +3,14 @@ import { ServiceUsageName } from '../google-cloud/types';
 import { IntegrationConfig } from '../types';
 import { collectEnabledServicesForProject } from './service-usage/client';
 
+const disabledServiceToStepMap: {
+  [key in ServiceUsageName]?: Set<string>;
+} = {};
+
+export function getDisabledServiceToStepMap() {
+  return disabledServiceToStepMap;
+}
```
ndowmon, line 12: Is there any way to make this a pure function? As a general rule, I try to avoid writing any stateful functions because they tend to be difficult to test/debug and can cause strange side-effects.

ndowmon:
I think this one has been superceded by https://github.com/JupiterOne/graph-google-cloud/pull/574. Can you see if this is still necessary, and close if not?

PR #559, opened 2022-11-28
Create data model for Web Security Scanner

Nick-NCSU:


docs/spec/src/steps/services/web-security-scanner/index.ts
```diff
@@ -0,0 +1,81 @@
...
+    name: 'Fetch Scan Configs',
+    entities: [
+      {
+        resourceName: 'Scan Config',
+        _type: 'google_cloud_scan_config',
+        _class: ['Configuration'],
+      },
+    ],
+    relationships: [],
+    dependsOn: [],
```
JakeFerrero, line 24: Not as familiar with the graph-google-cloud project, but just making sure, there's not service-level entity this would be dependent on?

docs/spec/src/steps/services/web-security-scanner/index.ts
```diff
@@ -0,0 +1,81 @@
...
+     * REQUIRED PERMISSIONS: cloudsecurityscanner.scanruns.list
+     *                       cloudsecurityscanner.crawledurls.list
+     */
+    id: 'fetch-scan-runs',
+    name: 'Fetch Scan Runs',
+    entities: [
+      {
+        resourceName: 'Scan Run',
+        _type: 'google_cloud_scan_run',
+        _class: ['Finding'],
```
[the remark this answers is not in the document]
JakeFerrero: I don't really have a better alternative. If no one else has a complaint, it's probably fine.

JakeFerrero (approved):

docs/spec/src/steps/services/web-security-scanner/index.ts
Nick-NCSU: Not that I am aware of, I am also not too familiar but from what I could find it didn't need any

PR #563, opened 2022-12-12
INT-5687 Limit when we print missing_permission job log for org policy

JakeFerrero:
When getting the public access prevention policy, only print the missing_permission job log when the customer is actually missing the orgpolicy.policy.get permission. Any other error returned should not result in that job log.


JakeFerrero:
To check what message was received when the `orgpolicy.policy.get` permission was missing, I created two service accounts - one with the permission and one without. I then made the `/policies/storage.publicAccessPrevention` API call like we do in the integration. Verified that the first one received a 403, and the second a 200.

gastonyelmini (approved):
Looks good to me!

i5o (approved):

PR #568, opened 2023-01-09
feat(INT-6599): create document permissions script

gastonyelmini:
### Context and motivation: 

We must provide the required list of permissions that require a review of all the SDK calls that we are doing and collect all of them. As part of making this process more scalable, this spike ticket was created, in order to investigate a way to automate the process of collecting this permissions and adding this to `jupiterone.md` README automatically. This script might be migrated to the SDK in the future to apply this process to all services. 

### How it works

`yarn document:permissions` command was added. This script imports `invocationConfig` variable from steps index to be able to collect permissions. This permissions will be added to the `IntegrationStep` array in each step as permissions. This would look like this:

```
{
    id: ...,
    name: ...,
    dependsOn: [],
    entities: [],
    relationships: [],
    executionHandler: () => {},
    permissions: ['permission1', 'permission2'],
},
```

And will produce a new section in `jupiterone.md` documentation:

Screen Shot 2023-01-09 at 11 01 20

### DIFF
--- a/commands/documentPermissions.ts
+++ b/commands/documentPermissions.ts
@@ -1,0 +1,107 @@
+import { promises as fs } from 'fs';
+import * as path from 'path';
+import { Command } from 'commander';
+import { invocationConfig } from '../src';
+import { GoogleCloudIntegrationStep } from '../src/types';
+
+const table = require('markdown-table');
+
+const documentPermissionsCommand = new Command();
+
+interface DocumentCommandArgs {
+  outputFile: string;
+}
+
+const J1_PERMISSIONS_DOCUMENTATION_MARKER_START =
+  '<!-- {J1_PERMISSIONS_DOCUMENTATION_MARKER_START} -->';
+const J1_PERMISSIONS_DOCUMENTATION_MARKER_END =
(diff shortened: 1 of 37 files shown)


ndowmon (dismissed):

ndowmon (approved):

zemberdotnet:
@ndowmon and @gastonyelmini This is seriously awesome. Is this something either of you think would be good to port upstream in time to the SDK? Documenting permissions is a tough task given the size of lots of integrations and I think this may help in other places as well. Thoughts?

gastonyelmini:
> @ndowmon and @gastonyelmini This is seriously awesome. Is this something either of you think would be good to port upstream in time to the SDK? Documenting permissions is a tough task given the size of lots of integrations and I think this may help in other places as well. Thoughts?

Hey @zemberdotnet, thank you for taking a look! I agree that this could be applied to the SDK to be used by other services, actually that is one of the considerations of this spike. I'm open to discuss at any time!

BTW, this is a chained pr: https://github.com/JupiterOne/graph-google-cloud/pull/570, that has some script improvements, a pre-commit hook, the actual permissions review, and the list of permissions added to the documentation.

PR #571, opened 2023-01-20
fix(INT-6706): check service account key in validateInvocationConfig

gastonyelmini:
### Context and motivation: 

In some cases, despite the fact that service account key file is setup correctly, if is not valid, the integration might run anyways and generate a large amount of errors. To avoid this, a test API call with credentials to check if they are valid and throw an error if not:

Example error that could be present:

Screen Shot 2023-01-19 at 23 11 25

Expected outcome: 

Screen Shot 2023-01-19 at 23 18 05


src/getStepStartStateExtraTests.test.ts
```diff
@@ -24,57 +19,46 @@ import { STEP_IAM_CUSTOM_ROLES, STEP_IAM_MANAGED_ROLES } from './steps/iam';
...
-        projectId: 'j1-gc-integration-dev-nested', // Project that doesn't have Cloud Asset API enabled and is in the same org as j1-gc-integration-dev-v3
-        // Temporary tweak to make this test pass since its recording has been updated from the new organization/v3
-        serviceAccountKeyFile: integrationConfig.serviceAccountKeyFile.replace(
-          'j1-gc-integration-dev-v2',
-          'j1-gc-integration-dev-v3',
-        ),
-        serviceAccountKeyConfig: {
-          ...integrationConfig.serviceAccountKeyConfig,
-          project_id: 'j1-gc-integration-dev-v3',
+    await withGoogleCloudRecording(
```
ndowmon, line 23: ❤️

src/getStepStartStateExtraTests.test.ts
```diff
@@ -24,57 +19,46 @@ import { STEP_IAM_CUSTOM_ROLES, STEP_IAM_MANAGED_ROLES } from './steps/iam';
...
-        disabled: false,
-      },
-      [STEP_IAM_MANAGED_ROLES]: {
-        disabled: false,
+      async () => {
+        const stepStartStates = await invocationConfig.getStepStartStates?.(
+          createMockExecutionContext({
+            instanceConfig: {
+              ...integrationConfig,
+              projectId: 'j1-gc-integration-dev-nested',
```
ndowmon, line 36: Why use `j1-gc-integration-dev-nested` instead of what's in your invocation config?

Or alternatively, could you update your `.env` to include `j1-gc-integration-dev-nested` when recording this test? This would help us avoid hard-coded configurations in tests.

src/config.test.ts
```diff
@@ -0,0 +1,77 @@
...
+  integrationConfig,
+} from '../test/config';
+import {
+  withGoogleCloudRecording,
+  getMatchRequestsBy,
+} from '../test/recording';
+import { invocationConfig } from './index';
+import { IntegrationConfig } from './types';
+
+describe('#config - validateInvocation tests', () => {
```
ndowmon, line 17: Let's do the following tests:

- success
- account not found (service account email is wrong - e.g. does not exist)
- account not found (project ID is wrong)
- expired access token

test/recording.ts
```diff
@@ -295,7 +295,10 @@ function redact(entry): void {
     // Remove recording binary marker
     delete (entry.response.content as any)._isBinary;
 
-    if (requestUrl === 'https://www.googleapis.com/oauth2/v4/token') {
+    if (
+      requestUrl === 'https://www.googleapis.com/oauth2/v4/token' &&
+      entry.response.status !== 400
+    ) {
```
gastonyelmini, line 301: I did this because error message was not being recorded correctly. Will redact access_token only if status is different from 400, because it does not include the token at all. `https://www.googleapis.com/oauth2/v4/token` is called before any other API call, to authenticate the client.

src/config.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  try {
+    await executeTestRequest(config);
+  } catch (err) {
+    if (err.status === 403) {
+      throw new IntegrationProviderAuthenticationError({
+        cause: err,
+        endpoint: 'serviceusage:services.list',
+        status: err.code,
+        statusText: `Google Cloud service account key is not allowed to access to the project specified in configurations. Please make sure that the project id of the integration configuration is correct.`,
```
gastonyelmini, line 32: Tried to add a more user friendly message if access to resource is forbidden. Because the actual message was a bit confusing because it was related to the actual test call. What do you think?

src/__recordings__/getStepStartStateKeyValidation_372270608/recording.har
```diff
@@ -0,0 +1,137 @@
...
+          "queryString": [],
+          "url": "https://www.googleapis.com/oauth2/v4/token"
+        },
+        "response": {
+          "bodySize": 181,
+          "content": {
+            "encoding": "utf-8",
+            "mimeType": "application/json; charset=UTF-8",
+            "size": 181,
+            "text": "{\"access_token\":\"[REDACTED]\",\"expires_in\":9999,\"token_type\":\"Bearer\"}"
```
ndowmon, line 70: I think this needs a re-record to persist the `400` response data.

src/index.ts
```diff
@@ -31,6 +32,7 @@ export const invocationConfig: IntegrationInvocationConfig<IntegrationConfig> =
     integrationSteps: steps,
     dependencyGraphOrder: ['last'],
     beforeAddEntity: maybeDefaultProjectIdOnEntity,
+    validateInvocation,
```
ndowmon, line 35: 💯

ndowmon (dismissed):
This is awesome. Thanks!

I posted the comment about re-recording one API call. If it’s simple to re-record, please do so, but if it takes more than a few minutes, i would rather merge as-is.

🔥

ndowmon (approved):

PR #575, opened 2023-02-07
fix(INT-7041): fix broken validate questions action

gastonyelmini:


Gonzalo-Avalos-Ribas (approved):
removed frozen lockfile

PR #576, opened 2023-02-20
fix(INT-6744): add debug log if timeout error

gastonyelmini:
## Add debug log if timeout error

Context: 

Google Cloud does not handle correctly errors when BitBucket credentials are expired. Added a log in order to make it more easy to debug in the future.

### DIFF
--- a/src/steps/cloud-build/client.ts
+++ b/src/steps/cloud-build/client.ts
@@ -1,7 +1,7 @@
 import { cloudbuild_v1, google } from 'googleapis';
 import { Client } from '../../google-cloud/client';
 import { IntegrationStepContext } from '../../types';
-import { CloudBuildLocations } from './constants';
+import { CloudBuildEntitiesSpec, CloudBuildLocations } from './constants';
 
 export class CloudBuildClient extends Client {
   private client = google.cloudbuild({ version: 'v1' });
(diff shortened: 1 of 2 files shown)


src/steps/cloud-build/steps/fetch-cloud-build-bb-repos.ts
```diff
@@ -40,22 +41,31 @@ export const fetchCloudBuildBitbucketRepositoriesStep: GoogleCloudIntegrationSte
...
+          } catch (err) {
+            if (err.code === 'ATTEMPT_TIMEOUT') {
+              logger.warn(
+                `${CloudBuildEntitiesSpec.BUILD_BITBUCKET_SERVER_CONFIG._type} - Unable to fetch BitBucket repositories. This might be caused by expired credentials in the GCP console (Cloud Build).`,
               );
-            },
-          );
+            }
+
+            throw err;
```
ndowmon, line 67: I think it would be better to try/catch inside `client.iterateBuildBitBucketRepositories()`. If we encounter `ATTEMPT_TIMEOUT` here, let's not throw (but throw if encountering any other errors).

src/steps/cloud-build/client.ts
```diff
@@ -129,29 +129,41 @@ export class CloudBuildClient extends Client {
...
+          for (const config of data.bitbucketServerRepositories || []) {
+            await callback(config);
+          }
+        },
+      );
+    } catch (err) {
+      if (err.code === 'ATTEMPT_TIMEOUT') {
+        context.logger.warn(
+          { err },
+          `${CloudBuildEntitiesSpec.BUILD_BITBUCKET_SERVER_CONFIG._type} - Unable to fetch BitBucket repositories. This might be caused by expired credentials in the GCP console (Cloud Build).`,
```
gastonyelmini, line 162: The name of the variable is `CloudBuildEntitiesSpec` but actually comes from a constants file

src/steps/cloud-build/client.ts
```diff
@@ -129,29 +129,41 @@ export class CloudBuildClient extends Client {
...
-        data: cloudbuild_v1.Schema$ListBitbucketServerRepositoriesResponse,
-      ) => {
-        for (const config of data.bitbucketServerRepositories || []) {
-          await callback(config);
-        }
-      },
-    );
+      }
+
+      throw err;
```
ndowmon, line 166: I suggest the following so that the step does not fail if `ATTEMPT_TIMEOUT` occurs.
```suggestion
      } else {
        throw err;
      }
```

ndowmon (approved):

PR #579, opened 2023-03-30
INT-7619: fix duplicated Audit Config IAM Policy relationship key

gastonyelmini:
#  Fix duplicated Audit Config IAM Policy relationship key

Google's `auditLogConfig.exemptedMembers` response is the following:

```
auditLogConfig [
  {
    logType: 'ADMIN_READ',
    exemptedMembers: [ 'user:gaston.yelmini@contractor.jupiterone.com' ]
  },
  {
    logType: 'DATA_READ',
    exemptedMembers: [ 'user:gaston.yelmini@contractor.jupiterone.com' ]
  }
]
```

This is generating a duplicated key error when a member is included in one or more actions. 

## Proposed solution

Merge relationships additionalProperties `logType ` and avoid adding it twice to the jobState.

### DIFF
--- a/src/steps/resource-manager/index.ts
+++ b/src/steps/resource-manager/index.ts
@@ -3,7 +3,6 @@ import {
   createDirectRelationship,
   createMappedRelationship,
   RelationshipDirection,
-  Relationship,
 } from '@jupiterone/integration-sdk-core';
 import { ResourceManagerClient } from './client';
 import {
@@ -55,6 +54,7 @@ import {
 } from '../service-usage/constants';
 import { getServiceApiEntityKey } from '../service-usage/converters';
 import { buildIamTargetRelationship } from '../cloud-asset';
+import { cloudresourcemanager_v3 } from 'googleapis';
 
 export * from './constants';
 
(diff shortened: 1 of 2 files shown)


src/steps/resource-manager/index.ts
```diff
@@ -340,12 +343,32 @@ export async function fetchIamPolicyAuditConfig(
...
           });
 
-          if (relationship) {
-            await jobState.addRelationship(relationship);
+          const foundIndex = iamTargetRelationships.findIndex(
+            (relationship) => relationship._key === relationship._key,
+          );
+
+          if (relationship && foundIndex === -1) {
+            iamTargetRelationships = [...iamTargetRelationships, relationship];
```
Gonzalo-Avalos-Ribas, line 351: Doing Array.concat might be faster than using the spread operator,specially for big arrays. Would you mind changing it? 
`iamTargetRelationships = iamTargetRelationships.concat(relationship);`
Should yield the same result!

gastonyelmini: Totally agree! Actually I would rather do just .push. Was reading at this article that does a benchmark of concat vs push and push seems to be faster: https://dev.to/uilicious/javascript-array-push-is-945x-faster-than-array-concat-1oki. Good catch!

src/steps/resource-manager/index.ts
```diff
@@ -340,12 +343,32 @@ export async function fetchIamPolicyAuditConfig(
...
+              ];
+            iamTargetRelationships[foundIndex].additionalProperties![
+              'logType'
+            ] = `${currentLogType}, ${logType}`;
           }
         }
       }
     }
+
+    await Promise.all(
```
ndowmon, line 367: I think implementing the following helper function would greatly simplify this step:


```ts

/**
 * Reorganizes from objects with an array of exemptedMembers and a single 
 * logType to a single exemptedMember and an array of logTypes
 * 
 * [
 *   { exemptedMembers: ['development@jupiterone.com'], logType: 'type1' }
 *   { exemptedMembers: ['development@jupiterone.com'], logType: 'type2' },
 * ]
 * 
 * => 
 * 
 * [
 *   { exemptedMember: ['development@jupiterone.com'], logTypes: ['type1', 'type2', ] }
 * ]
 * 
 */
function flattenAuditLogConfigs(
  auditLogConfigs: cloudresourcemanager_v3.Schema$AuditLogConfig
): {
  exemptedMember: string;
  logTypes: string[];
}[] {
  // todo implement
}
```

ndowmon: Then you should be able to do just this:

```ts
    for (const { exemptedMember, logTypes } of flattenAuditLogConfigs(auditConfig.auditLogConfigs || [])) {
      const parsedMember = parseIamMember(exemptedMember);
      const { identifier: parsedIdentifier, type: parsedMemberType } =
        parsedMember;
      let principalEntity: Entity | null = null;
      if (parsedIdentifier && parsedMemberType === 'serviceAccount') {
        principalEntity = await jobState.findEntity(parsedIdentifier);
      }
      const relationship = buildIamTargetRelationship({
        fromEntity: auditConfigEntity,
        principalEntity,
        parsedMember,
        logger,
        projectId: client.projectId,
        additionalProperties: { logTypes },
        relationshipClass: RelationshipClass.ALLOWS,
      });

      if (relationship) {
        await jobState.addRelationship(relationship);
      }
    }
```

ndowmon (approved):

PR #583, opened 2023-04-17
Consolidate event log names for better user experience

VDubber:
- prettier


src/utils/events.ts
```diff
@@ -12,7 +15,7 @@ export function publishMissingPermissionEvent({
   stepId,
 }: PublishMissingPermissionEventParams) {
   logger.publishEvent({
```
ndowmon, line 17: Should this be
```suggestion
  logger.publishWarnEvent({
```

zemberdotnet (approved):

PR #587, opened 2023-04-20
INT-7797: refactor private ca

gastonyelmini:
The issue is that we’re using privateca API version v1beta1, which is deprecated. We need to instead use v1, which is supported.

as a result, we need to also add the caPool resource, which should be a parent of google_certificate_authority.

This also includes:

- Migrating test to new version
- Refactoring service constants 
- Splitting handlers to multiple files
- Splitting steps that fetch and generate relationships in the same step.


src/steps/privateca/steps/buildCertificateAuthorityBucketRelationships.ts
```diff
@@ -0,0 +1,78 @@
...
+          },
+          'Could not find raw data on certificate authority entity',
+        );
+        return;
+      }
+
+      const bucketName = instance.gcsBucket;
+
+      if (!bucketName) {
+        return;
```
JuanManuelIbarlucea, line 42: Idem

src/steps/privateca/steps/buildCertificateAuthorityBucketRelationships.ts
```diff
@@ -0,0 +1,78 @@
...
+      }
+
+      const storageBucketEntity = await jobState.findEntity(
+        getCloudStorageBucketKey(bucketName),
+      );
+      if (!storageBucketEntity) {
+        return;
```
JuanManuelIbarlucea, line 49: Idem

src/steps/privateca/steps/buildCertificateAuthorityBucketRelationships.ts
```diff
@@ -0,0 +1,78 @@
...
+async function buildCertificateAuthorityBucketRelationships(
+  context: IntegrationStepContext,
+): Promise<void> {
+  const { jobState, logger } = context;
+
+  await jobState.iterateEntities(
+    { _type: PrivatecaEntities.PRIVATE_CA_CERTIFICATE_AUTHORITY._type },
+    async (caAuthorityEntity) => {
+      const instance =
+        getRawData<privateca_v1.Schema$CertificateAuthority>(caAuthorityEntity);
```
ndowmon, line 28: nit:
```suggestion
      const caAuthority =
        getRawData<privateca_v1.Schema$CertificateAuthority>(caAuthorityEntity);
```

src/steps/privateca/steps/buildCertificateAuthorityBucketRelationships.ts
ndowmon, line 1: It typically makes sense to put this step inside `fetchAuthorityCertificates.ts`, since the relationship is built off of that entity.

src/steps/privateca/steps/fetchAuthorityCertificates.test.ts
```diff
@@ -0,0 +1,40 @@
...
+import { invocationConfig } from '../../..';
+import { integrationConfig } from '../../../../test/config';
+import {
+  getMatchRequestsBy,
+  setupGoogleCloudRecording,
+} from '../../../../test/recording';
+import { PrivatecaSteps } from '../constants';
+
+describe(`privateca#${PrivatecaSteps.STEP_PRIVATE_CA_CERTIFICATES.id}`, () => {
+  let recording: Recording;
```
ndowmon, line 15: tip: this suggestion will prevent you from ever forgetting the condition in front of `if (recording) await recording.stop()`

```suggestion
  let recording: Recording | undefined;
```

JuanManuelIbarlucea (changes requested):
Great work! 
Just left some minor comments on possible logging.

src/steps/privateca/steps/fetchAuthorityCertificates.ts
```diff
@@ -0,0 +1,75 @@
...
+            data: certificate,
+            keyAlgorithm: certificateAuthorityEntity.keyAlgorithm as string,
+            projectId: client.projectId,
+          });
+
+          await jobState.addEntity(certificateEntity);
+
+          await jobState.addRelationship(
+            createDirectRelationship({
+              _class: RelationshipClass.CREATED,
```
ndowmon, line 53: Is this `CREATED` or `HAS`?

test/mocks.ts
```diff
@@ -1956,77 +1956,6 @@ export function getMockApiGatewayGateway(
   };
 }
 
-export function getMockCertificate(
```
ndowmon, line 1959: 💯

src/steps/privateca/steps/fetchCertificateAuthorities.ts
```diff
@@ -0,0 +1,107 @@
...
+    },
+    async (caPool) => {
+      const caPoolId = caPool._key.split('/')[5];
+      const location = caPool._key.split('/')[3];
+
+      if (!location || !caPool) {
+        if (!caPool) {
+          logger.warn(
+            `${PrivatecaSteps.STEP_PRIVATE_CA_CERTIFICATE_AUTHORITIES.id} - Unable to fetch CA due to missing parameters. location: ${location}, caPoolId: ${caPoolId}.`,
+          );
```
ndowmon: tip: 

```suggestion
          logger.warn({ location, caPoolId },
            `${PrivatecaSteps.STEP_PRIVATE_CA_CERTIFICATE_AUTHORITIES.id} - Unable to fetch CA due to missing parameters.`,
          );
```

src/steps/privateca/steps/fetchCaPools.ts
```diff
@@ -0,0 +1,41 @@
...
+    logger,
+  } = context;
+
+  const client = new PrivateCaClient({ config });
+
+  await client.iterateCaPools(async (caPool) => {
+    if (!caPool) {
+      logger.warn(
+        `${PrivatecaSteps.STEP_PRIVATE_CA_POOLS.id} - Found undefined caPool entity in iterateCaPools.`,
+      );
```
ndowmon, line 24: Is this something you've encountered? How is the API returning an array that contains`undefined`?

src/steps/privateca/steps/fetchAuthorityCertificates.ts
ndowmon: nvm - not a new relationship. You should keep as is!

src/steps/privateca/steps/fetchCaPools.ts
gastonyelmini: this was just for safe purposes but you are right it does not make much sense

src/steps/privateca/steps/fetchAuthorityCertificates.ts
```diff
@@ -61,6 +65,56 @@ async function fetchAuthorityCertificates(
...
+      }
+
+      const bucketName = caAuthority.gcsBucket;
+
+      if (!bucketName) {
+        logger.warn(
+          `${PrivatecaSteps.STEP_CREATE_PRIVATE_CA_CERTIFICATE_AUTHORITY_BUCKET_RELATIONSHIPS.id} - Missing bucketName.`,
+        );
+        return;
+      }
```
ndowmon, line 95: curious - is `gcsBucket` a required property here? Is it true that if we do not find it, something has gone wrong?

src/steps/privateca/steps/fetchAuthorityCertificates.test.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    const stepTestConfig: StepTestConfig = {
+      stepId: PrivatecaSteps.STEP_PRIVATE_CA_CERTIFICATES.id,
+      instanceConfig: integrationConfig,
+      invocationConfig: invocationConfig as any,
+    };
+
+    const result = await executeStepWithDependencies(stepTestConfig);
+    expect(result).toMatchStepMetadata(stepTestConfig);
+  });
```
ndowmon, line 39: Tip: instead of separately setting `jest.setTimeout(45_000)`, you can add the timeout as the third positional argument to `test()`:

```suggestion
  }, 45_000);
```

ndowmon (approved):

src/steps/privateca/steps/fetchAuthorityCertificates.ts
ndowmon:
```suggestion
      if (!bucketName) {
        return;
      }
```

JuanManuelIbarlucea (approved):
LGTM

PR #588, opened 2023-04-20
FIX(int-7914) Added error handling in case dataset/tables returns 404

Gonzalo-Avalos-Ribas:
Added error handling in case dataset/tables returns 404

### DIFF
--- a/src/steps/big-query/index.ts
+++ b/src/steps/big-query/index.ts
@@ -1,6 +1,8 @@
 import {
   createDirectRelationship,
   createMappedRelationship,
+  IntegrationLogger,
+  IntegrationProviderAPIError,
   RelationshipClass,
   RelationshipDirection,
 } from '@jupiterone/integration-sdk-core';
@@ -136,6 +138,7 @@ export async function fetchBigQueryModels(
   const {
     jobState,
     instance: { config },
+    logger,
   } = context;
   const client = new BigQueryClient({ config });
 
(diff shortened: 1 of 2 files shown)


src/steps/big-query/index.ts
```diff
@@ -145,24 +148,28 @@ export async function fetchBigQueryModels(
...
-            await jobState.addEntity(modelEntity);
+        try {
+          await client.iterateBigQueryModels(
+            datasetEntity.name as string,
+            async (model) => {
+              const modelEntity = createBigQueryModelEntity(
+                model,
+                datasetEntity.id as string,
+              );
+              await jobState.addEntity(modelEntity);
```
jroblesx, line 159: Can we add a validation to ensure that the entity was created before adding the relationship?

src/steps/big-query/index.ts
```diff
@@ -184,43 +191,67 @@ export async function fetchBigQueryTables(
...
-              createDirectRelationship({
-                _class: RelationshipClass.HAS,
-                from: datasetEntity,
-                to: tableEntity,
-              }),
-            );
-          },
-        );
+              await jobState.addEntity(tableEntity);
+              await jobState.addRelationship(
```
jroblesx, line 220: Can we add a validation to ensure that the entity was created before adding the relationship?

jroblesx (changes requested):
I added some comments, I don't know if its absolutely necessary but I feel myself more comfortable with a validation before adding the relationship.

ndowmon (approved):

src/steps/big-query/index.ts
ndowmon: I think this is out-of-scope for this change - this only has to do with error handling.

I also think that we can infer from the code that the entity was added before the relationship, so I don't think validation is necessary here.

ndowmon (approved):

jroblesx (approved):

PR #596, opened 2023-05-03
refactor(INT-7898): refactor Google Cloud Storage Bucket public property

gastonyelmini:
## Refactor Gooogle Cloud Storage Bucket public property.

This pr is a refactor on how we calculate is a bucket is public or not. This adds and `access` property that means the following: 

- `Public to internet` means one or more bucket-level permissions grant access to `allUsers` or `allAuthenticatedUsers`. 
- `Not public` means the bucket’s policy controls all objects uniformly, and no permissions have been granted to allUsers or allAuthenticatedUsers.
- `Subject to object ACLs` means fine-grained, object-level access control lists (ACLs) are enabled. Objects may be public if they grant access to allUsers or allAuthenticatedUsers.


src/steps/storage/index.ts
```diff
@@ -11,6 +11,117 @@ import {
...
+  return (
+    iamConfiguration?.uniformBucketLevelAccess?.enabled !== true &&
+    !publicPolicy
+  );
+}
+
+function getUsesEnforcedPublicAccessPrevention(
+  iamConfiguration: iamConfiguration,
+) {
+  return iamConfiguration?.publicAccessPrevention === 'enforced';
```
gastonyelmini, line 47: This actually changes how we calculate if is public or not , because the `Organization Public Access Prevention Policy` config might be `false` but if bucket's `publicAccessPrevention ` is `enforced ` this config is ignored. Additional explanation:

In Google Cloud Storage, publicAccessPrevention is a feature that helps prevent accidental exposure of data by blocking public access to buckets and objects. When you set the publicAccessPrevention field for a bucket, you can specify whether the bucket should inherit or enforce its public access prevention configuration.

Here's what each of these options means:

`Inherit`: When a bucket inherits its public access prevention configuration, it adopts the setting of its parent bucket or the project default setting. This means that if a parent bucket has public access prevention enabled, then any child buckets will also have it enabled by default. If the parent bucket doesn't have it enabled, then the child bucket won't have it enabled either.

`Enforce`: When a bucket enforces its public access prevention configuration, it requires that all objects in the bucket are either private or have appropriate access controls in place to prevent public access. This means that if you try to upload an object to a bucket that violates the bucket's public access prevention configuration, the upload will fail.

src/steps/app-engine/__snapshots__/index.test.ts.snap
```diff
@@ -47,7 +48,7 @@ Object {
       "kmsKeyName": undefined,
       "location": "EUROPE-NORTH1",
       "name": "dataproc-staging-europe-north1-167984947943-oqqo30p7",
-      "public": false,
+      "public": true,
```
gastonyelmini, line 51: This snap changes are because the services has old recordings. Will refactor this steps in other pr

PR #606, opened 2023-05-23
[INT-8168]: return google bigquery table size

mishelashala:


mishelashala:
After a review of the documentation and the typings from the library @ndowmon and I discovered the list endpoint doesn't return `numBtyes` nor `numRows`. Such properties are only returned by the individual `get` request. Some re-scoping needs to be done for this specific feature, closing PR in the meantime.

PR #614, opened 2023-07-10
INT-8173: expose cloudLoggingStatus

mishelashala:


austinkelleher (commented):
Can you explain why we switched this to be a string instead of a boolean? Thanks!

mishelashala:
@austinkelleher  I just reverted back the property to a `boolean`

Gonzalo-Avalos-Ribas (approved):
Lets make sure once this merges and the integration runs that we are actually getting false values

PR #618, opened 2023-07-25
INT-8705: avoid extra inventory calls if api is disabled

gastonyelmini:
Avoid making extra api calls if inventory api is disabled. Iterating thought a large amount of compute instances and trying to fetch instance inventory with the API disabled generates alerts in the dashboard.

### DIFF
--- a/src/steps/compute/index.ts
+++ b/src/steps/compute/index.ts
@@ -733,17 +733,24 @@ export async function fetchComputeInstances(
     compute_v1.Schema$ServiceAccount[]
   >();
 
+  let inventoryApiDisabled = false;
+
   await client.iterateComputeInstances(async (computeInstance) => {
     let instanceInventory: osconfig_v1.Schema$Inventory | undefined;
 
     try {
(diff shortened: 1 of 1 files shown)


src/steps/compute/index.ts
```diff
@@ -733,17 +733,24 @@ export async function fetchComputeInstances(
...
   >();
 
+  let inventoryApiDisabled = false;
+
   await client.iterateComputeInstances(async (computeInstance) => {
     let instanceInventory: osconfig_v1.Schema$Inventory | undefined;
 
     try {
-      if (computeInstance.zone) {
+      if (computeInstance.zone && !inventoryApiDisabled) {
```
mishelashala, line 742: since this is a "fix", can we add a unit test?

mishelashala (changes requested):

src/steps/compute/index.ts
gastonyelmini: I didn't find a good way to test the sdk implementation but will try.

PR #628, opened 2023-11-22
INT-10036: implement retry logic for attempt timeouts (NON api timeouts)

gastonyelmini:


src/google-cloud/client.ts
```diff
@@ -156,17 +168,84 @@ export class Client {
...
+                if (timeoutsRetryRemaining === 0) {
+                  throw new IntegrationProviderAPIError({
+                    code: RETRY_TIMEOUT_OPERATION_TIMEOUT_ERROR_CODE,
+                    status: 408,
+                    statusText: 'Operation Timeout',
+                    endpoint: 'UNKNOWN',
+                  });
+                }
+              } else {
+                return await this.withErrorHandling(fn, options);
```
Gonzalo-Avalos-Ribas, line 244: if resolveWithTimeout fails, the catch condition, has this if. In the case that is not a RETRY_TIMEOUT_OPERATION_TIMEOUT_ERROR_CODE this causes an infinite loop, because it calls again withErrorHandling, right?

PR #631, opened 2023-12-05
INT-8952: update SDK packages

gastonyelmini:


src/steps/compute/converters.ts
```diff
@@ -195,10 +195,7 @@ export function createComputeDiskEntity(
 
   return createGoogleCloudIntegrationEntity(data, {
     entityData: {
-      source: {
-        ...data,
-        tags: buildComputeInstanceTags(data),
-      },
+      source: data,
```
gastonyelmini, line 198: Deleted this because `data` (Disk) never had tags and was incorrectly set, only compute instance has tags. This appeared after googleapis package upgrade. You can check in snapshot tests that previous value of disk tags was undefined.

PR #688, opened 2024-05-20
INT-10878: update integration error handling

gastonyelmini:


src/google-cloud/client.ts
```diff
@@ -166,10 +166,10 @@ export class Client {
...
 
-    if (
-      (err.response?.status === 403 || err.status === 403) &&
-      !isQuotaLimitError(err)
-    ) {
+    const status =
+      err.response?.status !== undefined ? err.response?.status : err.status;
+
+    if ((status == 403 || status == 404) && !isQuotaLimitError(err)) {
       isAuthorizationError = true;
```
Gonzalo-Avalos-Ribas, line 173: 404 are usually not permissions issues, but i understand google sends it that way, can we add a comment about that?

Gonzalo-Avalos-Ribas (approved):

PR #696, opened 2024-06-24
set billing error as warning in KMS key rings step

RonaldEAM:


src/steps/kms/index.ts
```diff
@@ -28,25 +29,41 @@ export async function fetchKmsKeyRings(
...
+  } catch (err) {
+    if (
+      err.status === 400 &&
+      err.statusText?.match &&
+      err.statusText.match(/billing/i)
+    ) {
+      logger.publishWarnEvent({
+        name: IntegrationWarnEventName.IncompleteData,
+        description:
+          'Billing not enabled for the project. Skipping KMS Key Rings ingestion.',
```
gastonyelmini, line 61: Do we want this only for this step ? If not, we can catch this in the general error handling logic.

